PR #159, opened 2015-06-13
Hide Non Spam Option

shawniverson:
Another nice option, very useful if everything is being quarantined, not just spam.


shawniverson:
Missing a ; ... fixing and creating new pull request

PR #160, opened 2015-06-13
Hide Non Spam Option

shawniverson:
A nice option when all email is quarantined, including non-spam.


tools/Cron_jobs/quarantine_report.php
```diff
@@ -219,7 +219,13 @@
      ishighspam=0
     AND
      ishighmcp=0";
-}  
+}
+
+if (defined('HIDE_NON_SPAM') && HIDE_NON_SPAM) {
```
endelwar: please add ` === true` after `&& HIDE_NON_SPAM`

shawniverson: Done.  I will go back and update other options with this as well for consistency later.

PR #220, opened 2015-07-16
Fix for when ishighmcp is null and strict conditional for HIDE_HIGH_SPAM

shawniverson:


endelwar:
closed by 24938a5b93b8a888bd8f0bf110ccfcd269af6268

PR #308, opened 2016-10-21
Fixes #307

### LINKED ISSUE #307: Adding to whitelist/blacklist does not work when using german language
If you use german language and try to add addresses to the blacklist or whitelist this does nothing.

Skywalker-11:
The problem was that the value for the add button to add a whitelist or blacklist entry was used when determining what the action is the user requested. 
Due to the call of htmlentities at line 56 the "Hinzuf\xc3\xbcgen" from the input was converted to "Hinzuf&uuml;gen" (with html encoded entity für "ü"). 

At line 165 the value was then compared with the language specific "add07" value (here "Hinzuf\xc3\xbcgen"). This fails and so the adding to the list was not executed.

The second commit adds some missing translation to the german language file.

---

This change is [Reviewable](https://reviewable.io/reviews/mailwatch/1.2.0/308)


stefaweb:
Hello Skywalker-11!

For de.php, please use the last version in develop.

Sincerely,

Skywalker-11:
But theres the updated de.php from the develop branch.

Sorry for pushing to the wrong branch if needed I can also create a new clean pull request for the two files.

stefaweb:
Hello!

May be the best is to close this push and do a new one directly on develop.

Skywalker-11:
create clean pull request

PR #317, opened 2016-11-04
Fixes #307 and updates de-language file

### LINKED ISSUE #307: Adding to whitelist/blacklist does not work when using german language

Skywalker-11:
The problem was that the value for the add button to add a whitelist or blacklist entry was used when determining what the action is the user requested.
Due to the call of htmlentities at line 56 the "Hinzuf\xc3\xbcgen" from the input was converted to "Hinzufügen" (with html encoded entity für "ü").

At line 165 the value was then compared with the language specific "add07" value (here "Hinzuf\xc3\xbcgen"). This fails and so the adding to the list was not executed.

The second commit adds some missing translation to the german language file.

---
This change is [Reviewable](https://reviewable.io/reviews/mailwatch/1.2.0/317)


endelwar:
Thanks!

PR #324, opened 2016-11-27
Adjust login field width

shawniverson:
Login pane appears skewed after a recent commit.

---
This change is [Reviewable](https://reviewable.io/reviews/mailwatch/1.2.0/324)


endelwar:
Shouldn't it be better to enlarge input field to 300px, to match left and right padding? Or are you seeing other imperfection?

shawniverson:
That would also work.  I noticed that the original width was 308px.  Is there a reason it is now wider?  If so, that would make more sense.

stefaweb:
Hello!
Sorry I'm busy but can you check with fr.php and de.php.

endelwar:
It was enlarged for multi-language strings, some of them are longer that the English ones, see the French one down there.

schermata 2016-11-28 alle 12 43 59

shawniverson:
Ok, I will revise PR to focus on the input boxes instead :)

shawniverson:
Done, PR ready for review

PR #384, opened 2017-01-11
Upgraded Requests for PHP to v1.7

### LINKED ISSUE #378: LIB: Requests for PHP
We currently use version 1.6.1-MailWatch.

Version 1.7 is available.

https://github.com/rmccue/Requests/releases

asuweb:
Fixes #378

---
This change is [Reviewable](https://reviewable.io/reviews/mailwatch/1.2.0/384)


endelwar:
SOCKS5 support is missing, it has been added by me in 1.6.1-MailWatch version (https://github.com/mailwatch/1.2.0/commit/27a3a1401718b24bb7e2db4c020b15696d51e7cb)

It should be ported and tested with an actual SOCKS5 proxy before upgrading.

Note that 1.7 doesn't add or fix anything related to maxmind geoip database download.

asuweb:
Yes, good point!  I hadn't spotted that if I'm honest.

In which case, as you point out since there have been no functional changes, lets leave it as it is.

Issue #386, opened 2017-01-12
Graph no longer works: imageantialias() is not available


stefaweb:
All graph scripts no longer works on my new dev server.

I got:

`JpGraph Error: 25128 The function imageantialias() is not available in your PHP installation. Use the GD version that comes with PHP and not the standalone version.`

Debian Jessie
php5-gd 5.6.29+dfsg-0+deb8u1

/etc/php5/cli/conf.d/20-gd.ini,
GD Support => enabled
GD headers Version => 2.1.1-dev
gd.jpeg_ignore_warning => 0 => 0

stefaweb:
See lib/jpgraph/src/gd_image.inc.php

            JpGraphError::RaiseL(25128);//('The function imageantialias() is not available in your PHP installation. Use the GD version that comes with PHP and not the standalone version.')

stefaweb:
In case of need, found this fork of JpGraph.

https://github.com/umanit/jpgraph

stefaweb:
They also have a new 4.0.2 version.
http://jpgraph.net/download/

stefaweb:
Works if we comment out line 110 of gd_image.inc.php.

stefaweb:
Better solve.

In lib/jpgraph/src/jpg-config.inc.php at line 137 (at the end), add:

```
if(!ANTIALIASING){
    function imageantialias($image, $enabled){
        return true;
    }
}
```

endelwar:
you should report this also to JpGraph.

stefaweb:
I can do it if you want?

- update to 4.0.2
- patch for jpg-config.inc.php

endelwar:
I've looked better at this issues and seems related to Debian/Ubuntu decision to not include this function for their policy (see https://lists.debian.org/debian-user/2015/03/msg00283.html)

stefaweb:
Another resolution is to place this at the ned of  jpg-config.inc.php if Debian PHP.

`define('ANTIALIASING', false);`

No more.

stefaweb:
Updated PR.

```
// Patch for Debian PHP
define('ANTIALIASING', false);

if(!ANTIALIASING){
    function imageantialias($image, $enabled){
        return true;
    }
}
// End
```

Just for information. PHP 5.6 Dotdeb is not available for Debian Jessie. 
Only PHP version 7.

stefaweb:
Fixed in #389.

Skywalker-11:
[...]

PR #389, opened 2017-01-13
jpgraph 4.0.2 upgrade and patch for imageantialias()

stefaweb:
My first test using different "branch".

---
This change is [Reviewable](https://reviewable.io/reviews/mailwatch/1.2.0/389)


stefaweb:
See #386

endelwar:
Please remove examples direcotiry, they are not required and make the install package bigger

endelwar:
I think that is better to alter line 110 of `gd_image.inc.php` to `return true` instead of `JpGraphError::RaiseL(25128);` when this function is not implemented by linux distributions: a less defined graph is better that a error message.

Also for Debian is better to use dotdeb.org packages, which are more updated and contains the PHP forked GD library that contains this function (and others), Ubuntu people can use ondrej/php PPA

stefaweb:
I'm generally using dotdeb too but I was trying with "standard" Debian stuff first...

For jpgraph patch. Do as you prefer.

When the patch will be ok, I'll try a last time and will switch my server to dotdeb.

PR #403, opened 2017-01-18
Fixes uninitialised string offset

### LINKED ISSUE #394: Bug in reports.php (Add when empty)
If we click on Add without have done anything. We got an error.

Stay with the error until we click on "Remove".

capture d ecran 2017-01-14 a 16 26 20

asuweb:
Fixes #394

---
This change is [Reviewable](https://reviewable.io/reviews/mailwatch/1.2.0/403)

### DIFF
--- a/mailscanner/filter.inc.php
+++ b/mailscanner/filter.inc.php
@@ -232,7 +232,7 @@ WHERE
                         $sql .= "AND\n $val[0] $val[1]\n";
                     } else {
                         // Allow !<sql_function>
-                        if ($val[2]{0} === '!') {
+                        if ($val[2]!='' && $val[2]{0} === '!') {
                             $sql .= "AND\n $val[0] $val[1] " . substr($val[2], 1) . "\n";
                         } else {
                             // Regular string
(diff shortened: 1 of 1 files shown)


mailscanner/filter.inc.php
```diff
@@ -232,7 +232,7 @@ public function CreateSQL()
                         $sql .= "AND\n $val[0] $val[1]\n";
                     } else {
                         // Allow !<sql_function>
-                        if ($val[2]{0} === '!') {
+                        if ($val[2]!='' && $val[2]{0} === '!') {
```
endelwar: use strict comparation `!==` and check also with `isset`

asuweb: Agree with the strict comparison.

Not sure I agree with checking  `$val[2]`  with isset.  In this case, `$val[2]` will always be "set" by the post data from the form, so isset will always return true in this case.

endelwar: yes, you're right, just the strict comparison

stefaweb:
Can you forbid to click on Add if form "value" is empty?

mailscanner/filter.inc.php
stefaweb: Can you forbid to click on Add if form "value" is empty?

asuweb:
Updated with strict comparison

Issue #400, opened 2017-01-18
PDF


yogurtnaturalny:
Hi Mates 

I could not find "suggestion" tab so I will put this here

It is possible to add in version 1.3 option to create a pdf/print style report about spam ( when you go to see details about message )

I'm not so good with php but maybe you know how to pull this off ? 

Big thx for all your work and I look forward to hearing from you

Skywalker-11:
Would this be ok? Then I can make a patch
![zwischenablage01](https://cloud.githubusercontent.com/assets/11180862/22055428/6130a4c6-dd5b-11e6-8d57-374159b1099f.jpg)

Just need to add following css and apply `class="noprint"` to all parts that shouldn't be displayed when printing. For this page its 2 elements (logo and navigation bar).
```
@media print {
        .noprint {
                display: none;
        }
        .TD.heading-w175 {
                background: transparent;
        }
}
```

yogurtnaturalny:
This is what I was looking for, mate. A small patch but huge change :-D

endelwar:
merged with #407

PR #407, opened 2017-01-19
added css class to hide elements for printing (#400)

Skywalker-11:
This change is [Reviewable](https://reviewable.io/reviews/mailwatch/1.2.0/407)

### DIFF
--- a/mailscanner/functions.php
+++ b/mailscanner/functions.php
@@ -280,7 +280,7 @@ function html_start($title, $refresh = 0, $cacheable = true, $report = false)
     echo '</head>' . "\n";
     echo '<body onload="updateClock(); setInterval(\'updateClock()\', 1000 )">' . "\n";
     echo '<table border="0" cellpadding="5" width="100%">' . "\n";
-    echo '<tr>' . "\n";
+    echo '<tr class="noprint">' . "\n";
     echo '<td>' . "\n";
     echo '<table border="0" cellpadding="0" cellspacing="0">' . "\n";
     echo '<tr>' . "\n";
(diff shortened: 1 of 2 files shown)


yogurtnaturalny:
Hi, I would like to know when this option will be added to your awesome UI, please?

endelwar:
Hi, it was added on 19 Jan 2017, so it is present from release 1.2.0-rc5 and above

yogurtnaturalny:
Hi, thank you for that, maybe it is my fault but I can't get it right??? When I'm on email detail page and press Ctrl - P for printing this still not show up as was here #400 Am I doing something wrong?

Skywalker-11:
@yogurtnaturalny
Which version of MailWatch do you use? Can you give a screenshot of the page?

yogurtnaturalny:
MailWatch 1.2.2 
And apologise, I can't provide  screenshot, however, I'm on normal page https://mymailscanserver.com/detail.php?token=XXXX and this one I would like to print as report of spam to provide to 3rd party anti-spam system ( so we all can benefit  from it :-) )

yogurtnaturalny:
Hello

I think I found a problem, FireFox is very special and got special needs as Chrome works well, thank you for have a look into this and add this feature to your UI

Skywalker-11:
Which firefox version do you use? This what i get in firefox 53.0.3 on Windows for printing:
image

yogurtnaturalny:
It was Debian 8 FF 45 so maybe that why :-( apologies

Skywalker-11:
Some actions and links were not hidden when printing. The PR corrects this.

PR #410, opened 2017-01-19
Modified lists.php for more details in Audit Log

stefaweb:
And new version 0.3.8 for languages files.

---
This change is [Reviewable](https://reviewable.io/reviews/mailwatch/1.2.0/410)

PR #414, opened 2017-01-20
fixes #413

### LINKED ISSUE #413: Bug in Black / White List page
Seems many bugs in Black / White List page in  lists.php.

When we arrive on the page, the form is empty.

capture d ecran 2017-01-20 a 13 11 44

If we **create** an entry, after the refresh the form is not well emptied.

capture d ecran 2017-01-20 a 13 15 43

If we **delete** an entry, after the refresh the form is filled with some of the variable from the deleted entry.

21066080-df08-11e6-9951-fc74675f5200

Skywalker-11:
This change is [Reviewable](https://reviewable.io/reviews/mailwatch/1.2.0/414)

### DIFF
--- a/mailscanner/lists.php
+++ b/mailscanner/lists.php
@@ -228,6 +228,11 @@ if ($url_submit === 'delete') {
 
     $id = safe_value($url_id);
     dbquery($sql);
+
+   //empty the variables so that they aren't inserted in the input fields again
+    $to_domain = "";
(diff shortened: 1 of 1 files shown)


stefaweb:
Thanks. I modified PR #413 
Fix for radio button b/w and "Add" was missing too.

Skywalker-11:
it is in #410

PR #429, opened 2017-01-23
fixes #424

### LINKED ISSUE #424: Bug in lists.php (Go to message)
If we delete a black or white entry, we got this:

capture d ecran 2017-01-22 a 17 53 26

The number is the entry number ($id) of the deleted entry.

Skywalker-11:
fixes #424 

This change is [Reviewable](https://reviewable.io/reviews/mailwatch/1.2.0/429)

### DIFF
--- a/mailscanner/lists.php
+++ b/mailscanner/lists.php
@@ -64,7 +64,7 @@ $url_domain = (isset($_GET['domain']) ? sanitizeInput($_GET['domain']) : '');
 $url_domain = htmlentities($url_domain);
 $url_domain = safe_value($url_domain);
 
-$url_id = (isset($_GET['id']) ? sanitizeInput($_GET['id']) : '');
+$url_id = (isset($_GET['listid']) ? sanitizeInput($_GET['listid']) : '');
 $url_id = htmlentities($url_id);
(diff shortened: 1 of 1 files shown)


stefaweb:
delete no longer works
when we click nothing append

stefaweb:
delete of white list don't work
delete of black list works

stefaweb:
Seem the problem is more complex.
Sometime, the entry is deleted, sometime not.

Skywalker-11:
Works here. if you do a var_dump of `$_GET['listid']` does that contain a valid id for the blacklist/whitelist entry?

Skywalker-11:
There shouldn't be a difference between black/whitelist cause both use the same function to make the link and delete the entries

stefaweb:
Have a look of the screen shot.

I click to delete the entry 881 and after a short "refresh", the entry is still in the list.

capture-d ecran-2017-01-23-a-11 11 45

stefaweb:
Sorry, my lists.php file was not uptodate.
I replaced all the file and now it works.

Skywalker-11:
ready for merge

PR #466, opened 2017-02-02
New CRONs (Cron_jobs)

stefaweb:
Various changes in tools directory, CRONs and INSTALL doc for 1.2.0.

- created one uniq cron for mailwatch to be put in /etc/cron.daily
- created a cron for Postfix MTA to be put in /etc/cron.hourly
- moved all .php MailWatch system scripts in /usr/local/bin
- move out from /mailscanner dir postfix_relay.php and mailscanner_relay.php (security)
- renamed various files to make them easier to understand
- renamed Sendmail_queue directory to MailWatch_queue
- deleted obsolete files
- updated various INSTALL doc
- created entry in UPGRADING file for MailWatch 1.2.0


asuweb:
Can you combine all of the new cron scripts into a single file, that tests and then runs each script in turn?

Could you put a control file that points to the mailwatch functions.php in /usr/local/bin/ and then the php scripts can reference that file - it reduces the amount of user editing that is required, and reduces duplication?

The cron scripts should really declare which shell they're to be run in.

Skywalker-11:
I wouln't use /usr/local/bin because it usually is used for binary files only (like the name suggests).
How about /etc/mailwatch?

stefaweb:
> I wouln't use /usr/local/bin because it usually is used for binary files only (like the name suggests).
> How about /etc/mailwatch?

My wish will be to have the 3 php files directly moved to /var/www/html/mailscanner. 

db_clean.php
quarantine_maint.php
quarantine_report.php

One place! Easy to maintain. 😀

Personally, I don't like the idea to place "executable" php files in /etc/.

Skywalker-11:
I ment the control file that points to mailwatch directory. I agree with moving that files

Skywalker-11:
We also need localization in that files 😁

asuweb:
I agree, I wouldn't advocate putting the control file in /bin either.  I think the php scripts are fine there though. 

I wouldn't put the files in a web directory, as they're not intended to be executed via the web, and would be a fantastic place for a bot to attack / ddos.

stefaweb:
Zut!

Where do you want them?

endelwar:
> My wish will be to have the 3 php files directly moved to /var/www/html/mailscanner.

oh please no! :) script files *must* reside outside public web directory. They must be run from cli, not through web server/browser

endelwar:
In my installations the entire git repository is in `/opt/mailwatch` and I point my webserver root dir to `/opt/mailwatch/mailscanner` (so that I can have `https://mailwatch.mydomain.it`) and then I symlink cron files to /etc/cron.daily and symlink `MailWatch.pm` in `/usr/share/MailScanner/perl/custom`.

This way I can upgrade MailWatch with `git pull`

stefaweb:
Just let me now where?

/usr/local/bin  (Debian)
/usr/local (FHS)

asuweb:
I use a very similar setup to @endelwar - it makes upgrading that much easier.  That would by my choice as well, that was there nothing to move, only some symlinks which need creating, and no chance of things being missed on upgrade.

stefaweb:
Good discussion. ;)

> symlink cron

This has security implications. The scripts in /etc/cron.daily/ (and others) are run by root with root privileges. It's a bad idea to symlink scripts there that are owned and writable for non-root users, since your system could be compromised that way.

asuweb:
You can get round that by making sure the scripts can only be run and edited by root - I don't know how most people use mailwatch / mailscanner, but I use them on dedicated servers solely as a gateway, so no users ever need (or have access) to run anything.

asuweb:
Totally unrelated, but I've always been curious as to why the folder containing the "mailwatch" web interface is called "mailscanner" - I've always though calling it "mailwatch" or "web" or something would be more reflective of what it contains.

stefaweb:
Hackers love "special" built machine with nice hole.

An the database "mailscanner"...   ☺️

Skywalker-11:
Can you integrate [https://github.com/Skywalker-11/1.2.0/tree/feature/localizeQuarantineReport](https://github.com/Skywalker-11/1.2.0/tree/feature/localizeQuarantineReport). It fixes the plain text view for the report

endelwar:
FHS suggest that a package should stay in /opt or /srv: I think that /opt is a good place to put a webapp like MailWatch.
I've just filled a request to Linux Foundation to reserve `mailwatch` namespace for cron files and /opt/ directory (also init file if we need that in future): https://lsbbugs.linuxfoundation.org/show_bug.cgi?id=4196

endelwar:
> Totally unrelated, but I've always been curious as to why the folder containing the "mailwatch" web interface is called "mailscanner" - I've always though calling it "mailwatch" or "web" or something would be more reflective of what it contains.

Well, I don't know either: it has always been called this way by the original author Steve Freegard (hi @smfreegard ! ), or at least from version 0.3beta dated 5 September 2003.

I'd like to keep this naming convention for the 1.2 serie, V2 will have a db called mailwatch.

Skywalker-11:
Do we move mailwatch in 1.2.0 or in 2.0 to /opt?

endelwar:
> Do we move mailwatch in 1.2.0 or in 2.0 to /opt?

v2.0

stefaweb:
Need help for Postfix_relay directory to update INSTALL with more detail.

The files /var/www/html/mailscanner/postfix_relay.php and /var/www/html/mailscanner/mailscanner_relay.php should be moved to Postfix_relay directory and later placed in /usr/local/bin (as the php files in Cron_jobs).

stefaweb:
Question for Postfix lover.

In mailwatch_relay cron, we have:

```
test -x /usr/local/bin/postfix_relay.php || exit 0
/usr/bin/php -q /usr/local/bin/postfix_relay.php --refresh

test -x /usr/local/bin/mailscanner_relay.php || exit 0
/usr/bin/php -q /usr/local/bin/mailscanner_relay.php --refresh
```

Does this file output something when run?
Can we add ">/dev/null 2>&1" at the end?

Do you want a separate (and optional) cron file or do you want the commands placed in the regular "mailwatch" cron file but commented?

stefaweb:
Forgot my second question. It's stupid as "mailwatch_relay" goes in /etc/cron.hourly/.

Do you want "mailwatch" or "mailwatch_relay" as name?

shawniverson:
mailwatch_relay makes sense to me.  

These scripts typically don't output anything, unless something is broken.

stefaweb:
@endelwar: the PR is ready to merge but I don't know how to resolve the conflict with mailscanner/postfix_relay.php.

I moved the file from mailscanner/postfix_relay.php to tools/Postfix_relay.

stefaweb:
Seems I have again broken something...

stefaweb:
Created an new PR #487. Will close this one later.

endelwar:
> Do you want "mailwatch" or "mailwatch_relay" as name?

I prefer minus sign `-` over underscore `_`.
Also filenames should be prefixed by `mailwatch-`, followed by something significant and possibly short:

- /etc/cron.d/mailwatch-postfix-relay
- /etc/cron.d/mailwatch-sendmail-relay
- etc...

stefaweb:
I can do that.

PR #473, opened 2017-02-04
Fix for jpGraph in debian (missing imageantialias)

Skywalker-11:
fix for #386


stefaweb:
I tried. Works fine.

mailscanner/functions.php
```diff
@@ -72,7 +72,12 @@
 } else {
     $lang = require __DIR__ . '/languages/' . LANG . '.php';
 }
-
+if (!defined('USE_ANTIALIASING') || !USE_ANTIALIASING) {
```
endelwar: just change this line with the following and the new config entry is not needed:
```php
if (!function_exists('imageantialias')) {
```

Skywalker-11: done

Issue #463, opened 2017-01-31
PHP Preg error in functions.php - MailWatch can't find the virus scanner in MailScanner configuration when set to auto


GoranLundberg:
The following errors appear in /var/log/apache2/error.log every time MailWatch is loaded.

```
[Tue Jan 31 19:54:43.336264 2017] [:error] [pid 12503] [client 192.168.1.1:49302] PHP Notice:  Use of undefined constant VIRUS_REGEX - assumed 'VIRUS_REGEX' in /var/www/mailscanner/functions.php on line 1112
[Tue Jan 31 19:54:43.336554 2017] [:error] [pid 12503] [client 192.168.1.1:49302] PHP Warning:  preg_match(): Delimiter must not be alphanumeric or backslash in /var/www/mailscanner/functions.php on line 1112
```
And nothing is showing up under "Recent Messages" in the MailWatch interface.

**Edit:** Had a faulty solution to the problem that i removed.

**Solution:**
The problem is that MailWatch does not parse the configuration item `Virus Scanners` correctly in the MailScanner configuration file when set to `auto`. On Debian it's located in `/etc/MailScanner/conf.d/yourlocal.conf` (Replace yourlocal with the name of your custom configuration file)

I changed `Virus Scanners = auto` to `Virus Scanners = clamd`
and now it seems to be working fine.

This problem should be fixed and handled correctly in MailWatch.
Hope someone can make a fix so that no one else has to debug it.
Took several hours searching the web trying to fix this.

Thanks @DiederikvandenB for mentioning the solution!

Skywalker-11:
What MailWatch and anti virus module are you using?

Normally the constant VIRUS_REGEX should be defined as soon as the functions.php is loaded (line 120-178). The only case in which the constant is not set (that's what the PHP Notice says) is when you are using an unsupported anti virus.

DiederikvandenB:
I am experiencing the exact same issue. Possibly interesting to note that I am running PHP7.0, not 100% sure whether that's an influencing factor here but I thought it was worth noting. Furthermore I am running the master version of MW and ClamAV as anti-virus.

DiederikvandenB:
I seemed to have fixed the issue myself already. I had the `Virus Scanner` setting in MailScanner set to `auto`. This is not recognised by MailWatch, causing the error. Changing this setting to `clamd`fixed the issue, however I feel like MW should handle this setting appropriately (no offense to the developers obviously, thank you for making the software ;) )

[...]

PR #475, opened 2017-02-04
possible fix for auto/unknown virus scanner

Skywalker-11:
This should prevent error messages like #463 where an unknown virus scanner is defined or 'auto'.
It will show an info message in the report table in the top right part of the page (not shown to normal users)

### DIFF
--- a/mailscanner/languages/en.php
+++ b/mailscanner/languages/en.php
@@ -185,6 +185,7 @@ return array(
     '15minutes03' => '15 min.:',
     'saspam03' => 'Spam',
     'sanotspam03' => 'Not spam',
+    'unknownvirusscanner03' => 'Unknown virus scanner',
 
     // 04-detail.php
     'receivedon04' => 'Received on:',
--- a/mailscanner/functions.php
+++ b/mailscanner/functions.php
@@ -1156,9 +1156,9 @@ AND
     arsort($virus_array);
     reset($virus_array);
     // Get the topmost entry from the array
-    if(!defined('VIRUS_REGEX')) {
+    if (!defined('VIRUS_REGEX')) {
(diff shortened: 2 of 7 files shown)


stefaweb:
> 'unknownvirusscanner03' => 'Unknown virus scanner',

'unknownvirusscanner03' => 'Logiciel anti virus inconnu',

mailscanner/languages/en.php
```diff
@@ -185,6 +185,7 @@
     '15minutes03' => '15 min.:',
     'saspam03' => 'Spam',
     'sanotspam03' => 'Not spam',
+    'unknownvirusscanner03' => 'Unknown virus scanner',
```
endelwar: This should be more specific, like: 
Unknown virus scanner defined in MailScanner.conf, review your configuration and don't use 'auto' as antivirus name (see FAQ)

asuweb:
@Skywalker-11 - does this also fix the issue with esets being the first virus scanner defined?

stefaweb:
For fr.php:

`'unknownvirusscanner03' => 'Le scanner de virus défini dans MailScanner.conf est inconnu. Modifiez votre configuration et n\'utilisez pas \'auto\' comme nom d\'antivirus (voir FAQ).',`

Skywalker-11:
@asuweb This should prevent broken pages but mailwatch still only uses the first virus scanner set in mailscanner.conf. So yes it should prevent the PHP Notice and PHP Warning with esets but can't use the virus data.

asuweb:
@Skywalker-11 - great :)

stefaweb:
Localization in #492 but DE needed to do for 'unknownvirusscanner03'.

PR #482, opened 2017-02-05
Updated disclaimer in all .awk files

stefaweb:
One day, .awk files will need to be localized (or find a solution to no longer use awk).


stefaweb:
Ready to merge.

endelwar:
One day awk files will go away!

PR #484, opened 2017-02-05
fix #480

### LINKED ISSUE #480: BUG in reports.php (double list of reports when changing language)
When we change of language, we got the report list in double. One set per language.

We have to quit the browser and launch it again to go back to the normal.

capture d ecran 2017-02-05 a 09 56 07

Skywalker-11:
maybe we should consider to remove that part of generating the report list out of the filter (doesn't realy make sense)


endelwar:
I agree, this doesn't make sense at all

Issue #453, opened 2017-01-29
Wishlist: Install script for Mailscanner / Mailwatch


asuweb:
A lot of the mailwatch installation process could be fairly easily scripted - having just done 2 parallel installs, there's very little that couldn't be handled automatically with the relevant arguments supplied to the script.

This would likely reduce the number of failed setups and idiosyncrasy which makes debugging particularly tricky.  

It could also be set to automatically fetch and install mailscanner should the user wish

endelwar:
I agree with you! I was thinking the same last week.
And this should be shipped with 1.2.0, and be very well tested before release.

- create.sql has to be replaced by an install.php console script
- upgrade.php has to be included in install.php which has to check tables existence, field length and encoding, the it should be removed
- perl libraries version should be checked
- documentation has to be updated together with code
- blame systemd for its existence ;)

stefaweb:
I hate systemd too !

Good idea install script.

asuweb:
@endelwar - Great minds think alike.  Shouldn't you be changing nappies? ;-)

Yes, it definitely needs extensive testing.  

I was thinking more a shell script rather than php, which would handle the installation and setup of the full stack, that way there's some control over what's installed and incompatibilities can be minimised.

Skywalker-11:
And we have to decide what operating systems we want to support. We won't be able to make the install script work for every system. Also if we put the instructions for all operating systems in one big file or split them for each os. 

I suggest that we split the upgrade/install into multiple files (sql, package/perl management, file copying / permissions). That way they can easier be maintained.

As for supported systems I suggest:
Webservers: nginx, Apache (with virtual-host or conf.d file; maybe specify the web folder path for MailWatch during setup)
OS: Debian, Ubuntu, SUSE

asuweb:
Centos, RHEL

stefaweb:
Hello!

Write install scripts and test them is very time consuming.

If possible, I will prefer that we publish RC5 "as this" with the curent incoming PR before switching to a new install concept.

asuweb:
I agree writing and maintaining install scripts is time consuming, but then so if find "bugs" which may or may not bugs in the concept of mailwatch.   If we can standardise the install / setup process, then there should be far less time consumed tracking down issues which seem to be system / setup related.

stefaweb:
Like #-4-3-0!

stefaweb:
This tool have an interesting system to manage OS.

[...]

PR #485, opened 2017-02-05
Apache config files for MailWatch

stefaweb:
Basic Apache configuration files and doc.


stefaweb:
Do we need "short_open_tag = On"?
I don't remember.

Skywalker-11:
we always use `<?php` to start php part so there shouldn't be a need for short_open_tag

stefaweb:
Ready to merge.

asuweb:
@stefaweb - these files are specific to Debian based systems, and are not compatible with other *nix systems, such as red-hat based variants.

stefaweb:
I don't have RedHat .rpm server.
If you have one, I can add the config and doc. 
And add directory in tools/Apache for Debian, RedHad and others if needed.

Skywalker-11:
@asuweb do you mean the mod_php part? If that's the case we may should split apache and php configuration. 

@stefaweb we should add a `Require all-granted`  for apache 2.4 or for version 2.2 
```
Order allow,deny
Allow from all
```

asuweb:
@Skywalker-11 - sorry, no the actual config will probably work, but the instructions won't.  

Do we want to make this a global config or a vhost?  Personally I use a vhost for the mailwatch interface, rather than a global config.

The proposed install script (#453) should probably take care of this aspect as well

asuweb:
For redhat based systems (redhat / centos / fedora)

The apache dirs are 
 `/etc/httpd/`
`/etc/httpd/conf.d/`

The `a2en...` commands do not exist

`service httpd restart`

endelwar:
This should belong to https://github.com/mailwatch/mailwatch-docs, in a dedicated section that talks about server configuration and other tips related to MailWatch

stefaweb:
I agree. There are dozens of ways to do it. I can not generate vhost configs for all systems and way to do.

Better to close this PR.

stefaweb:
Close this. 
Better in MailWatch online doc.

PR #507, opened 2017-02-08
fixes #346 : removes old entries from the list before adding new ones

### LINKED ISSUE #346: Bug with whitelist/blacklist
With last DEV version (RC4+).

When I delete an entry in the whitelist, the entry is still used by mailscanner until I restart it with "service mailscanner restart".

Same when I add an entry in whitelist.

Same with blacklist.

Skywalker-11:
This fixes #346. Problem was that we keep the blacklist/whitelist till it gets refreshed. When that happens we query all entries from sql and add them to the old list which still contains the old entries and may overrides them. If an entry was removed from the sql db in the mean time it remains in the blacklist/whitelist for the perl skript till it process is restarted (normally restart of mailscanner).

To fix this all entries are removed from the list when it gets updated.

To see that behavior you can include Dumper via `use Data::Dumper` and insert the following in the sub CreateList before the `$sql = "SELECT to....` and after the second while loop
`MailScanner::Log::WarnLog("DEBUG: %s", Dumper($BlackWhite));`

This will print `DEBUG: $VAR1 = {};` if the lists are empty and eg. the following when an entry exists to the mail.log:
```
MailScanner[1467]:           $VAR1 = {
MailScanner[1467]:           'test.example.org' => {
MailScanner[1467]:                                  'bla@example.org' => 1
MailScanner[1467]:                                }
MailScanner[1467]:         };
```

I'm still new to perl so tell me if I missed something


stefaweb:
@asuweb 

If you have some spare time!

Could you add these counters?

capture d ecran 2017-02-09 a 11 39 15

stefaweb:
Sorry not at the right place.

Skywalker-11:
applied fixes for pm files to prevent possible errors while testing

stefaweb:
Good idea to merge all this.
Very verbose when you have 8000 blacklist entries... But good for real testing.

stefaweb:
@Skywalker-11 :   add a notice close to refresh time variable.

Do go below 5 minutes.

;-)

stefaweb:
> enables debug output (remove it when merging)

Comment them only.

Skywalker-11:
Maybe it is useful to do more output eg. the rule that caused black/whitelisting

stefaweb:
Good idea!

PR #419, opened 2017-01-21
Password reset functionality

asuweb:
Provides password reset functionality for users. See issue #139 

Minimal base code changes for optimal integration.  

Language file translations at en.php ` //100 - `password_reset.php` to ease maintainability until merged.  Should be re-ordered once merged.

---
This change is [Reviewable](https://reviewable.io/reviews/mailwatch/1.2.0/419)

PR #519, opened 2017-02-13
Fixes #518

### LINKED ISSUE #518: Quarantine_report - incorrect paths
quarantine_report.php has undefined variable `$mailwatch_dir`

Image hardcoded, should use `MW_LOGO`

asuweb:
Fixes #518


mailscanner/conf.php.example
```diff
@@ -91,7 +91,7 @@ define('MS_CONFIG_DIR', '/etc/MailScanner/');
 define('MS_SHARE_DIR', '/usr/share/MailScanner/'); // path for ConfigDefs.pl file
 define('MS_LIB_DIR', '/usr/lib/MailScanner/'); // path for sophos-wrapper file
 define('MS_EXECUTABLE_PATH', '/usr/sbin/MailScanner');
-define('IMAGES_DIR', './images/'); // trailing slash needed
+define('IMAGES_DIR', '/images/'); // trailing slash needed
 define('CACHE_DIR', './images/cache/'); // JpGraph cache
```
stefaweb: And these?

'./'

define('CACHE_DIR', './images/cache/'); // JpGraph cache
define('TTF_DIR', './lib/jpgraph/src/fonts/'); // JpGraph fonts

asuweb: I left those in as I hadn't the time to find out why they were like that in the first place?  

The remaining leading (.) don't appear to be causing any bugs.

stefaweb: OK

Skywalker-11:
The patch can cause problems if mailwatch is not in the root dir of the web seever but eg. www.exaple.org/mailwatch in which case the browser would look in invalid directory www.example.org/images for the logo instead of www.exaple.org/mailwatch/images

mailscanner/conf.php.example
stefaweb: @asuweb corrected this in #419

asuweb: @stefaweb - I thought I'd done it already!  Good spot :)

asuweb:
This comes back to different setups again then.  

It's not an issue with the use of virtual hosts or when in the web root..  Perhaps it should be fixed as follows:

`define('MAILWATCH_WEB_ROOT' , '/');`
`define('IMAGES_DIR', '.images/'); // trailing slash needed`

`' . __('mailwatchtitle03') . '`

Skywalker-11:
Yes that would be the most compatible solution

asuweb:
I'll do it that way then :)

Skywalker-11:
On second thought the mailwatch_quarantine_report is not shown on the website and therefor the fix would not work for that issue. Can't we just use MAILWATCH_HOME to include the pear libraries? No need to change IMAGES_DIR.

stefaweb:
@asuweb 

Could you pease add spaces before and after the MW logo?

capture d ecran 2017-02-14 a 20 48 32

stefaweb:
On line 495, hard code logo.

`$mime->addHTMLImage(MAILWATCH_HOME . '/images/mailwatch-logo.png', 'image/png', 'mailwatch-logo.png', true);`

asuweb:
@stefaweb - that's fixed in the refactoring in password-reset PR.  I moved the email sending to functions.php and replaced the hard coded paths

mailscanner/conf.php.example
endelwar: won't change this.

endelwar:
> The patch can cause problems if mailwatch is not in the root dir of the web seever

Backward compatibility has to be maintained

asuweb:
@endelwar - yes I agree.  I'm going to re-think this.

Skywalker-11:
Can you tell in which case something won't work the way it is now (with `./images`)? it is only used here
```
./mailscanner/auto-release.php:    <img src="<?php echo MAILWATCH_HOSTURL . IMAGES_DIR . MW_LOGO; ?>" alt="<?php echo __('mwlogo99'); ?>">
./mailscanner/conf.php.example:define('IMAGES_DIR', './images/'); // trailing slash needed
./mailscanner/functions.php:    echo '<td align="left"><a href="index.php" class="logo"><img src="' . IMAGES_DIR . MW_LOGO . '" alt="' . __('mailwatchtitle03') . '"></a></td>' . "\n";
./mailscanner/functions.php:        'IMAGES_DIR',
./mailscanner/login.php:    <div style="text-align: center"><img src="<?php echo IMAGES_DIR . MW_LOGO; ?>" alt="<?php echo __('mwlogo99'); ?>">
```

asuweb:
@Skywalker-11 - it seems to work everywhere except quarantine_report - but I'll check that just now with the latest dev branch

asuweb:
This fixes the quarantine images issue, and also the path issue as reported in #551

Skywalker-11:
@asuweb it works for me with 
`$mime->addHTMLImage(MAILWATCH_HOME . '/./images/mailwatch-logo.png', 'image/png', 'mailwatch-logo.png', true);`  
and when using the defined constants.
`$mime->addHTMLImage(MAILWATCH_HOME . '/' . IMAGES_DIR . MS_LOGO, 'image/png', 'mailwatch-logo.png', true);
`

The '/' is important here

asuweb:
@Skywalker-11 - yes, you're quite correct.  Don't know why I didn't try that the first time!

I've done that in the latest commit

PR #521, opened 2017-02-14
Remove last MS_LOGO in code

stefaweb:


mailscanner/functions.php
```diff
@@ -3946,7 +3946,6 @@ function checkConfVariables()
         'MS_EXECUTABLE_PATH',
         'MS_LIB_DIR',
         'MS_LOG',
-        'MS_LOGO',
```
endelwar: Add it to `$obsolete`

stefaweb: done
I had not seen that array. ;-)

PR #547, opened 2017-02-17
Update Queue commands for Exim

stefaweb:
If Exim is used as MTA, we use `exim -bpc` instead of call to the database.

Added entry for Exim in `/tools/sudo/mailwatch`.


stefaweb:
Do we need to put `exim -bpc` and `exim -bpc -DOUTGOING` in conf.php.example?

mailscanner/functions.php
```diff
@@ -426,12 +426,16 @@ function html_start($title, $refresh = 0, $cacheable = true, $report = false)
...
+                // Else use mailq which is for Sendmail and Exim
             } elseif (MAILQ) {
-                $inq = database::mysqli_result(dbquery('SELECT COUNT(*) FROM inq WHERE ' . $_SESSION['global_filter']),
-                    0);
-                $outq = database::mysqli_result(dbquery('SELECT COUNT(*) FROM outq WHERE ' . $_SESSION['global_filter']),
-                    0);
+                // If Exim, use dedicated command, else use database
+               if ($mta === 'exim') {
+                   $inq = exec('sudo /usr/sbin/exim -bpc 2>&1');
+                   $outq = exec('sudo /usr/sbin/exim -bpc -DOUTGOING 2>&1');
```
endelwar: is `OUTGOING` traversal to all exim installation or is this queue name configurable?

stefaweb: Default for Debian/Ubuntu but not for other system.
Can be configured in the Exim config files but if the user already have a running config, it may be better to have this in conf.php for custom.

As:

define('EXIM_EXEC_INQ', '/usr/sbin/exim -bpc');
define('EXIM_EXEC_OUTQ', '/usr/sbin/exim -bpc -DOUTGOING');

stefaweb: I dont' know if equivalent command exist for Sendmail.

stefaweb: Found `sendmail -bp` but not just a number.

https://www.electrictoolbox.com/show-sendmail-mail-queue/

I'm not a pro of regex...

stefaweb: For Sendmail outgoing queue:

`sendmail -bp -OQueueDirectory=/var/spool/mqueue.in`

Skywalker-11:
Shouldn't the constants be optional? Eg. they are not needed when using postfix.
- EXIM_QUEUE_IN
- EXIM_QUEUE_OUT

The following code would make it optional
```
...
 if ($mta === 'exim') {
                   $inq = exec('sudo ' . (defined(EXIM_QUEUE_IN) ? EXIM_QUEUE_IN : '/usr/sbin/exim -bpc' ) . ' 2>&1');
                   $outq = exec('sudo ' .  (defined(EXIM_QUEUE_OUT) ? EXIM_QUEUE_IN : '/usr/sbin/exim -bpc -DOUTGOING' )  . ' 2>&1');
} else ...
               
```

stefaweb:
Seems a problem in function.php in develop. 
"optional lists" commented?

```
    /*
    // TODO: Implement optional lists
    $optional = array(
        'RPC_PORT',
        'RPC_SSL',
        'VIRUS_REGEX',
        'LDAP_BIND_PREFIX',
        'LDAP_BIND_SUFFIX',
    );
    */
```

Skywalker-11:
They just aren't checked anywhere. Because they are optional that is not a problem

Issue #549, opened 2017-02-17
LDAP configuration "too" tight


remkolodder:
Hi,

We use a distributed setup for our mailscanner environment. Our webserver runs the MailWatch application, allowing our users to login and do their thing with the application and messages as they please.

We have a non-standard configuration, where our fields are different then the normal fields. instead of having our users defined as cn=user... we have the variable uid=user as main selector for our users.

The email field contains the email address of the user, but this is basically the same as the 'uid' field.
So for us to get this working we changed the LDAP_EMAIL_FIELD to uid and hit our first problem.

When the LDAP_EMAIL_FIELD does not match "email" the configuration forces the field to be "SMTP:$user" but this "field" is not available in our configuration at all (none of our schema's have this). Can the 

if (LDAP_EMAIL_FIELD === 'email' && strpos($user, '@')) {
            $ldap_search_results = ldap_search($ds, LDAP_DN, LDAP_EMAIL_FIELD . "=$user") or die(__('ldpaauth203'));
} elseif (strpos($user, '@')) {
            $ldap_search_results = ldap_search($ds, LDAP_DN, LDAP_EMAIL_FIELD . "=SMTP:$user") or die(__('ldpaauth203'));
        } else {

be made optional or "configurable" via conf.php so that we do not have to modify this code? We currently change the 'elseif' to match the if $ldap_search_result because that works for us. Why is the SMTP:$user being looked up btw? (e.g. where does it come from?)

In addition the 

 } else {
                    // build DN for user, when LDAP server is not an AD
                    $user = 'cn=' . $result[0]['cn'][0] . ',' . LDAP_DN;
                }

line, assumes that a user is defined as 'cn', but for us that is not the case. We can imagine that others also have different fields that markup their users. Can this also be made configurable via conf.php so that people with a different scheme can use the settings as  well without changing the functions.php (risking overwriting the changes everytime we update, note that we try to follow the develop branch as close as possible, yes at our own risk :-)).

Ofcourse, we accept the fact that we might run a non-default configuration and that we fall out of the boat for this one, if that is the case we need to be aware for every update that we revert these fields.

Skywalker-11:
@remkolodder can you check #552 if it would be sufficient?

stefaweb:
Is this issue solved?

remkolodder:
Yes this issue is solved. I will close the issue when submitting this comment.

[...]

PR #552, opened 2017-02-17
Generalize ldap for more flexible use

Skywalker-11:
More generalized version for ldap authentication. Should make use cases like #549 possible by configuration (no code change needed).

### DIFF
--- a/mailscanner/functions.php
+++ b/mailscanner/functions.php
@@ -2628,15 +2628,7 @@ function ldap_authenticate($user, $password)
         }
 
         //search for $user in LDAP directory
-        if (LDAP_EMAIL_FIELD === 'mail' && strpos($user, '@')) {
-            $ldap_search_results = ldap_search($ds, LDAP_DN, LDAP_EMAIL_FIELD . "=$user") or die(__('ldpaauth203'));
-        } elseif (strpos($user, '@')) {
-            $ldap_search_results = ldap_search($ds, LDAP_DN,
-                LDAP_EMAIL_FIELD . "=SMTP:$user") or die(__('ldpaauth203'));
-        } else {
-            // Windows LDAP (with legacy NT support)
-            $ldap_search_results = ldap_search($ds, LDAP_DN, "sAMAccountName=$user") or die(__('ldpaauth203'));
-        }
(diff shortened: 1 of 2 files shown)


remkolodder:
I think this might be sufficient, but the current proposed usage is only in conf.php, while I think we should modify functions.php as well to actually use it?

I also spotted a little typo, will comment on that.

mailscanner/functions.php
```diff
@@ -2677,7 +2675,7 @@ function ldap_authenticate($user, $password)
                     return null;
                 }
 
-                $bindResult = @ldap_bind($ds, $user, $password);
+                $bindResult = @ldap_bind($ds, $userlogin, $password);
```
remkolodder: $userlogin does not exist, this should remain $user .

mailscanner/functions.php
```diff
@@ -2664,11 +2656,17 @@ function ldap_authenticate($user, $password)
...
+                      !isset($result[0][LDAP_USERNAME_FIELD][0])) {
+                    @trigger_error(__('ldapno03') . ' "' . LDAP_USERNAME_FIELD . '" ' . __('ldapresults03'));
+                    return null;
+                }
+
+                if (LDAP_USERNAME_FIELD === 'cn') {
                     // build DN for user, when LDAP server is not an AD
                     $user = 'cn=' . $result[0]['cn'][0] . ',' . LDAP_DN;
+                } else {
+                    $user = $result[0][LDAP_USERNAME_FIELD][0];
```
remkolodder: This way, the $result will just be the field filled in, but a search needs the
complete DN.

for me this is needed;

> $user = LDAP_USERNAME_FIELD . '=' . $result[0][LDAP_USERNAME_FIELD][0] . ',' . LDAP_DN;

mailscanner/functions.php
Skywalker-11: done

mailscanner/functions.php
Skywalker-11: Introduced prefix and suffix because to enable to do this.

endelwar:
Very good!
Some config examples should be provided in docs for major user case (Active Directory or openldap standard installations)

PR #553, opened 2017-02-18
Insert default `admin` and fix schema in create.sql 

stefaweb:
Better to create admin by default.

No problem with SpamSetting with it.


endelwar:
I don't think that adding a default admin user with a default password is a great move: ever heard about Mirai botnet, where devices with default credential (admin/admin, admin/password and so on) get powned and used to ddos attacks?

A lazy sysadmin that leave the default credentials leave the entire email privacy to anyone not so kind to pass over it...

stefaweb:
But if the user forgot to create an administrator "admin", the SQL SpamSetting script will not work as needed.

endelwar:
> But if the user forgot to create an administrator "admin"

Lazy sysadmin that don't read all of INSTALL doc?

stefaweb:
> Lazy sysadmin that don't read all of INSTALL doc?

Yes, like me.... I replaced admin by another user name...

We can may be add a routine to check if the user `admin` still use `admin` as password. And ask the user to change it.

asuweb:
What's the purpose of the default admin?

endelwar:
on line 260 of SQLSpamSettings.pm the `admin` user default spam score can be returned.
I think that it's better to add a variable in `MailWatch-DB.pm` that ask for default admin username

stefaweb:
> What's the purpose of the default admin?

A creasy coding in SQLSpamSettings.pm. If admin user not present, spam setting do not work.

```
# Based on the address it is going to, choose the correct Spam score.
# If the actual "To:" user is not found, then use the domain defaults
# as supplied by the domain administrator. If there is no domain default
# then fallback to the system default as defined in the "admin" user.
# If the user has not supplied a value and the domain administrator has
# not supplied a value and the system administrator has not supplied a
# value, then return 999 which will effectively let everything through
# and nothing will be considered Spam.
#
```

stefaweb:
> I think that it's better to add a variable in MailWatch-DB.pm that ask for default admin username

I've done this in a .pm script I've wrote for another project. The easiest solution.

In fact, two variables. One for low score and one for high score.

asuweb:
Yes, I think a user configurable value would be better than adding a default admin account

stefaweb:
Ok, I remove the admin user from this PR but publish it to correct the schema for float.

We will do another PR to handle spamscore, highspamscore, noscan in MailWatch-DB.pm

endelwar:
> but publish it to correct the schema for float

`noscan` is a boolean variable, no need to set it to float

Skywalker-11:
like I said its working here without that admin user

stefaweb:
> like I said its working here without that admin user

For me.
No spam check at all unless for the one defined in user table.
= no default value for others -> spam goes to the users

stefaweb:
Can be closed. Float are in develop.

Skywalker-11:
I just noticed that new users don't have highspamscore and spamscore set. Can that be the cause its not working like you expected it?

stefaweb:
@Skywalker-11 

A problem in the GUI?

I'm not sure to understand the story of "0= default" in the GUI.

But the best, is to have the default values in MailWatch-DB.pm also.

Skywalker-11:
just noticed that it is related to change to float in the db and its default value which should be 0 and not NULL

stefaweb:
Do a PR for that.

PR #558, opened 2017-02-20
Fixes pear conflict

### LINKED ISSUE #395: Conflict with system PEAR packages
Having PEAR installed via yum or apt create conflict on local MailWatch PEAR packages.

`Fatal error: Cannot redeclare _PEAR_call_destructors() (previously declared in /var/www/html/mailscanner/lib/pear/PEAR.php:773) in /usr/share/pear/PEAR.php on line 777`

reported on #392 and in mailing list

asuweb:
This PR fixes #395 and allows the use of system installed PEAR.

@endelwar - This PR may conflict with #519 due the path change in mailwatch_quarantine_report.php (in PR #519), as the same section of code is modified in this PR.


stefaweb:
conflict

asuweb:
Yes, I thought it would - I'll fix

stefaweb:
Conflict.
What this PR is still here?

endelwar:
It's not merged yet because this PR solves the "fatal error: cannot redeclare" error, but system pear packages are different from the ones distributed with MailWatch

stefaweb:
conflict

Skywalker-11:
Do we want to update the PEAR core files? The most recent stable version we could use is 1.9.5 (from 2014). Newer versions are incompatible with PHP5.3 which still is used on CentOS 6 for example.

shawniverson:
@Skywalker-11  Eww, a rotten PEAR.  Yuck.  I am in favor of bumping to 1.9.5.  I think PHP 5.3 support should keep going a little longer while CentOS 6 is in its Maintenance Updates part of its lifecycle, though.

endelwar:
In December 2016 PEAR was updated to 1.10.1 (see https://github.com/mailwatch/MailWatch/commit/c4c06fa47bd9ae3450a36c19519268e695bfdcf3) along with other pear packages, but there are some editing to original pear files, like editing PEAR Net Socket to be php5 compatible or silencing some pear::error.

What would be good for this PR to be closed is finding a safe way to ignore completely system installed pear and use only pear shipped with MailWatch

endelwar:
On develop branch PEAR packages are loaded via composer, so there should be no more conflict with system installed PEAR

PR #561, opened 2017-02-20
Replaces function net_match with ip_in_range to use IPSet

### LINKED ISSUE #560: IPv6 and private network in functions.php?
Not urgent but seems some codes in functions.php have hard IP coded routine not compatible IPv6 and not using the new IPSet.php library.

asuweb:
Replaces function net_match with ip_in_range to use IPSet
Removes hard coded IPv4 private address space
Fixes #560


shawniverson:
w00t ipv6 for the win!

PR #562, opened 2017-02-21
DISPLAY_VIRUS option on message listing

stefaweb:
Some Virus definition ara very very long and breaks message listing view. 
Make this an option (true by default).
Grouped VIRUS constant in conf.php.example.

Sample of very long signature.
capture d ecran 2017-02-21 a 09 13 17


mailscanner/conf.php.example
```diff
@@ -174,6 +174,12 @@ define('QUARANTINE_SENDMAIL_PATH', '/usr/sbin/sendmail');
 // about a given virus or virus alias.  Comment out or set to false to disable.
 define('VIRUS_INFO', 'http://www.securelist.com/en/descriptions?words=%s&amp;behavior=&amp;Search=Search&amp;search_type=1');
 
+// Display Virus on message listing
+define('DISPLAY_VIRUS', true);
```
endelwar: could you please rename this to `DISPLAY_VIRUS_REPORT`?

PR #563, opened 2017-02-21
Provides required field validation for user_manager

asuweb:
Adds required fields for username & password
Provides better style for error display


stefaweb:
For fr.php

```
'erroruserreq12' => 'Nom d\'utilisateur nécessaire,
'errorpwdreq12' => 'Mot de passe nécessaire',
```

stefaweb:
Found this:

`echo ' <TR><TD CLASS="heading">' . __('retypepassword12') . '</TD><TD><INPUT TYPE="PASSWORD" ID="retypepassword" NAME="password1"></TD></TR>'`

May be `NAME="retypepassword"` will be better?

asuweb:
It works as it is, and it's been like that for years.  I can't really see any benefit in changing it

stefaweb:
All this have been modified one month ago by @Skywalker-11 and me when we have localized and corrected some bugs there. The second name change for the 2 password fields have been forgotten. But it's just cosmetic.

PR #569, opened 2017-02-22
Moved MailWatch perl file DB settings to separate file 

### LINKED ISSUE #568: Move DB settings for perl files to separate config file
Move the DB settings in each .pm file to a separate file to ease installation and upgrading.

asuweb:
Will make installation and updating easier by removing the need to changing db settings in .pm files

Fixes #568

### DIFF
--- a/MailScanner_perl_scripts/MailWatch-DB.pm
+++ b/MailScanner_perl_scripts/MailWatch-DB.pm
@@ -1,0 +1,6 @@
+#!/usr/bin/perl
+sub get_db_name { return 'mailscanner'};
+sub get_db_host { return 'localhost' };
+sub get_db_user { return 'mailwatch' };
+sub get_db_password { return 'mailwatch' };
+1;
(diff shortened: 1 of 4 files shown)


endelwar:
@asuweb this is great!
is it possible (I'm not a perl monk) to extract `return 'value'` in sub and put all of them as variables on top of the file, like this (pseudo code follow):
```perl
#!/usr/bin/perl

#  Modify this as necessary for your configuration  
my ($db_name) = 'mailscanner';
my ($db_host) = 'localhost';
my ($db_user) = 'mailwatch';
my ($db_pass) = 'mailwatch';

###############################
# don't touch below this line #
###############################
sub mailwatch_get_db_host { return $db_host };
etc...
```

This way, I think, it's easier for non perl proficient sysadmin to edit the files, without messing with functions syntax.

asuweb:
Yes, in fact that's what I did when I tested it but for some reason I didn't do it on this PR.  Fixed in latest commit :)

PR #575, opened 2017-02-23
fix for #555

### LINKED ISSUE #555: Spam scores should be float in mysql db
In users table the spam scores should be floats to enable finer definition of spam scores

```
spamscore` float DEFAULT NULL,
  `highspamscore` float DEFAULT NULL,
  `noscan` float DEFAULT NULL,
```

Skywalker-11:
fixes #555 default value


Skywalker-11:
double of #574

Issue #534, opened 2017-02-15
mailwatch_postfix_relay.php - bug


asuweb:
mailwatch_postix_relay.php no longer logs to the database.

From a review of the code it appears that the function ` extractKeyValuePairs($match)` is not called - at least I can't find reference to it. 

@Skywalker-11 - I think you re-wrote this?  Any ideas?

stefaweb:
Nothing is missing.

From RC4.

```
mailscanner/mtalogprocessor.inc.php:    abstract public function extractKeyValuePairs($match);
mailscanner/mtalogprocessor.inc.php:                $this->entries = $this->extractKeyValuePairs($match);
mailscanner/postfix_relay.php:    public function extractKeyValuePairs($match)
tools/Sendmail_relay/sendmail_relay.php:    public function extractKeyValuePairs($match)
```

From DEV:

ailscanner/mtalogprocessor.inc.php:    abstract public function extractKeyValuePairs($match);
mailscanner/mtalogprocessor.inc.php:                $this->entries = $this->extractKeyValuePairs($match);
tools/Sendmail_relay/mailwatch_sendmail_relay.php:    public function extractKeyValuePairs($match)
tools/Postfix_relay/mailwatch_postfix_relay.php:    public function extractKeyValuePairs($match)

Have you look if this something related to the move of the file from /mailscanner to /usr/local/bin.

asuweb:
Yes, the bit I can't find is the call to the `parse()` function within mtalogprocessor.inc.php`

asuweb:
Fixed in PR #535

Skywalker-11:
The extractKeyValuePairs is called by the super class [mtalogprocessor.inc.php](https://github.com/mailwatch/1.2.0/blob/develop/mailscanner/mtalogprocessor.inc.php#L178)

stefaweb:
Hello!
Just to understand. Why do we have the same function in 3 different files and not used in 2?

asuweb:
@Skywalker-11 - thanks, I found it.  

The issue was we were not explicitly stating the fields to be updated in the SQL statement, and a possible schema change may have stopped the query running due to column mis-match.  Took a while to debug as the `dbquery()` doesn't throw an error.

asuweb:
@stefaweb - the joy of object orientated programming.  

`public function extractKeyValuePairs($match)` these define the specific function for each of the MTA's, the

[...]

PR #576, opened 2017-02-23
possible fix for #534

Skywalker-11:


mailscanner/database.php
```diff
@@ -67,7 +67,12 @@ public static function connect($host = '', $username = '', $password = '', $data
      */
     public static function close()
     {
-        return self::$link->close();
+        $result = $true;
+        if (self::$link != null) {
```
endelwar: what about 
```php 
if (self::$link instanceof mysqli) { 
```
? (also for line 50)

mailscanner/database.php
```diff
@@ -32,7 +32,7 @@
 class database
 {
     /** @var mysqli $link */
-    public static $link;
+    public static $link = null;
```
endelwar: class propriety default value is already NULL if not specified

Skywalker-11: done

mailscanner/database.php
Skywalker-11: done

mailscanner/database.php
```diff
@@ -47,7 +47,7 @@ private function __construct()
      */
     public static function connect($host = '', $username = '', $password = '', $database = '')
     {
-        if (self::$link == null || !self::$link) {
+        if (self::$link instanceof mysqli) {
```
endelwar: this should be negative:
`if (!self::$link instanceof mysqli) {`
`instanceof` has precedence over `!` as operator

Skywalker-11: done

PR #588, opened 2017-02-23
Update comment for MAILWATCH_FROM_ADDR in conf.php.example

stefaweb:


mailscanner/conf.php.example
```diff
@@ -145,7 +145,8 @@ define('SHOW_MORE_INFO_ON_REPORT_GRAPH', false);
 /* Mailwatch Interface Settings */
 define('MAILWATCH_MAIL_HOST', '127.0.0.1');
 define('MAILWATCH_MAIL_PORT', '25');
-define('MAILWATCH_FROM_ADDR', 'postmaster');
+// Change for your own email address
```
endelwar: maybe `Change with a fully qualified email address` reads better

asuweb: Change it to postmaster@domain.tld then there is less confusion

endelwar: postmaster@domain.tld if fine for me, it's the commented description that can be confusing: to I have to put my very own email address or a generic corporate email like postmaster@domain.tld?

stefaweb: Or?

define('MAILWATCH_FROM_ADDR', 'postmaster@' . rtrim(gethostname()) . '');

asuweb: No, it should be a valid email address, otherwise it might end up in spam.... 

There's no guarantee that that would yield a valid email address.

endelwar: not always `gethostname()` return a FQDN.

Issue #595, opened 2017-02-25
$refresh_time in SQLSpamSettings.pm


stefaweb:
$refresh_time and the associated codes are not present in SQLSpamSettings as we have in SQLBlackWhiteList.pm.

Is this normal? 

If not, it will may be interesting to add it.

stefaweb:
Done in #597.

PR #597, opened 2017-02-26
Add refresh time to SQLSpamSettings.pm (#595)

stefaweb:
Its done if needed.
Tested on my dev server. No crash and refresh works.


stefaweb:
May be "MailWatch-DB.pm" will be better as "MailWatch-conf.pm"?

PR #611, opened 2017-03-07
fixes #609

### LINKED ISSUE #609: Bug in viewmail.php
I have this error, when I view a message (a dangerous content, message part).

capture d ecran 2017-03-07 a 08 59 23

capture d ecran 2017-03-07 a 09 01 12

Skywalker-11:


remkolodder:
This also goes haywire when you run into a GPG/PGP signed message. Seems like an endless loop.

PR #623, opened 2017-03-09
Checks relay ip's are only read from Received header

### LINKED ISSUE #622: Bug with IPSet.php when wrong IP in detail.php
I got an error when one of the IP is wrong.

```
Warning: inet_pton(): Unrecognized address 1489062189 in /var/www/html/mailscanner/lib/IPSet.php on line 200

Warning: inet_pton(): Unrecognized address 1489062189 in /var/www/html/mailscanner/lib/IPSet.php on line 200

Warning: gethostbyaddr(): Address is not a valid IPv4 or IPv6 address in /var/www/html/mailscanner/detail.php on line 158
```

capture d ecran 2017-03-09 a 13 28 05

Apparently the number comes from the header here:

```
X-Bluewin-Spam-Score: 0.00
X-FXIT-IP: IPv4[128.90.20.16] Epoch[1489062189]
```

asuweb:
Fixes #622

### DIFF
--- a/mailscanner/functions.php
+++ b/mailscanner/functions.php
@@ -2544,7 +2544,7 @@ function get_mail_relays($message_headers)
     $relays = null;
     foreach ($headers as $header) {
         $header = preg_replace('/IPv6\:/', '', $header);
-        if (preg_match_all('/Received:.+\[(?P<ip>.+)\]/', $header, $regs)) {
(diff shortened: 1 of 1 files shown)


endelwar:
why has the `if (false !== filter_var($relay, FILTER_VALIDATE_IP)) {` been removed in second commit?

asuweb:
I have no clue!  Brain error most likely!  I've re-added it

Issue #629, opened 2017-03-10
New symbol for message listing?


stefaweb:
What do you think to change the symbol [#] used in message listing?

In functions.php line 2068

![capture d ecran 2017-03-10 a 14 50 29](https://cloud.githubusercontent.com/assets/12907102/23797633/1fa27890-05a1-11e7-92fc-17281b849841.png)

This one is `&#9989;`

![capture d ecran 2017-03-10 a 14 50 44](https://cloud.githubusercontent.com/assets/12907102/23797641/24eb0574-05a1-11e7-890b-fc5c5c53525a.png)

Plenty of HTML symbol here: https://www.w3schools.com/charsets/ref_utf_dingbats.asp

asuweb:
HTML5 is not supported fully on older browsers.  Possibly better to use something like font awesome, for better compatibility.

My vote would be for: http://fontawesome.io/icon/info-circle/

stefaweb:
Can we have color for `info-circle`?

asuweb:
yes

asuweb:
You can do anything with them that you can in normal css - like size, color etc

stefaweb:
Ok for me. The "i" is easy to understand. You can do the change?

Just some lines before 2068 line, we have:

```
                case 'id2':
                    $fieldname[$f] = '#';
                    $orderable[$f] = false;
                    $align[$f] = 'center';
                    break;

```

Is this related to [#]?

asuweb:
@stefaweb - is this ok for you now?

stefaweb:
Yes, it works.

stefaweb:
Works fine. #631 can be merge.

stefaweb:
Done in #631.

PR #631, opened 2017-03-11
Add fa-info-circle to replace #

asuweb:
Fixes issue #629


endelwar:
I like the idea, but I don't like the full font-awesome inclusion.
I've made a svg version of the icon, with 2 png image fallback (blue and green) that can be used like this:
```html
<img src="info-circle.svg" onerror="this.src='info-circle.png'; this.onerror=null;">
```
[info-circle.zip](https://github.com/mailwatch/1.2.0/files/847585/info-circle.zip)

Can you work on these images?

asuweb:
Will do

asuweb:
@endelwar - I've changed it to use css background images instead as it was easier to apply hover effects

endelwar:
please remove the font files, so I can merge

asuweb:
Done :)

PR #640, opened 2017-03-13
Provides administrator the ability to combine quarantine reports into a single report

### LINKED ISSUE #608: Problem with Quarantine report and user/domain
Hello!

Something wrong with quarantine report.

USER user1@domain1.tld receive all message for domain1.tld

ADMIN DOMAIN user@domain.tld receive all domain1.tld but not message for domain.tld (domain1.tld is in filter for domain.tld)

Could you verify on your side?

asuweb:
Fixes #608 

Allows administrators to combine quarantine_reports into a single report containing all of the users filters.

Fixes a bug in the case of a single report per filter, where previously each filter report received quarantine mail for the parent user as well.

### DIFF
--- a/tools/Cron_jobs/mailwatch_quarantine_report.php
+++ b/tools/Cron_jobs/mailwatch_quarantine_report.php
@@ -347,7 +347,7 @@ ORDER BY a.date DESC, a.time DESC';
                     if (count($quarantined) > 0) {
                         $list = '';
                         foreach ($quarantine_list as $item) {
-                            $list .= $item . '<br>';
+                            $list .= $item . ', ';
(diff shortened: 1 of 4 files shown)


tools/Cron_jobs/mailwatch_quarantine_report.php
```diff
@@ -345,6 +345,7 @@
                         foreach ($quarantine_list as $item) {
                             $list .= $item . ', ';
                         }
+                        $list = substr($list,0,-2);
```
asuweb: Yes, I'm not sure why I didn't use that the first time! Doh!

PR #649, opened 2017-03-18
Adds RBL list to database, and displays relevant RBL lists in detail.php

### LINKED ISSUE #431: Wishlist: Show RBL List in Details Page
I would like that when MailScanner encounters a message listed in one or more RBL lists, these lists would be shown on the details page on the line "Listed in RBL" like below:

Listed in RBL: Y SPAMHAUS, BARRACUDA

screen

asuweb:
Fixes #431 

Adds the ability to display the list of RBL's that affect a message. 
Requires a very small modification MailScanner's `Message.pm`.

PR submitted to MailScanner https://github.com/MailScanner/v5/pull/45

Code can also be found here: https://github.com/asuweb/v5/commit/b9069a9f9124930c9cf2594824019c60fe56495a

image

### DIFF
--- a/upgrade.php
+++ b/upgrade.php
@@ -439,10 +439,19 @@ if ($link) {
 
     // Add new column and index to maillog table
     echo pad(' - Add maillog_id field, rblspamreport and primary key to `maillog` table');
-    if (true === check_column_exists('maillog', 'maillog_id', 'rblspamreport')) {
+    if (true === check_column_exists('maillog', 'maillog_id')) {
         echo color(' ALREADY DONE', 'lightgreen') . PHP_EOL;
     } else {
-        $sql = 'ALTER TABLE `maillog` ADD `maillog_id` BIGINT(20) UNSIGNED NOT NULL AUTO_INCREMENT, ADD PRIMARY KEY (`maillog_id`), ADD `rblspamreport` mediumtext COLLATE utf8_unicode_ci DEFAULT NULL';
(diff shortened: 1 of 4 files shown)


upgrade.php
```diff
@@ -438,11 +438,11 @@ function color($string, $color = '')
     }
 
     // Add new column and index to maillog table
-    echo pad(' - Add maillog_id field and primary key to `maillog` table');
-    if (true === check_column_exists('maillog', 'maillog_id')) {
+    echo pad(' - Add maillog_id field, rblspamreport and primary key to `maillog` table');
+    if (true === check_column_exists('maillog', 'maillog_id', 'rblspamreport')) {
```
endelwar: I'm pretty sure that this won't work (3 params?).
Also, in this case,  it's better to add single column: if migration from RC4 to RC5 has already been done this will not work

asuweb: Yes, good point

mmgomess:
Thanks a lot.

stefaweb:
No need of modification in MCPMessage.pm?

asuweb:
Not that can I can think of.  It's working on mine reliably as far as I can see.  I haven't specifically looked at MCPMessage.pm.

`rblspamreport` returns `rblspamheader` which was always returned to Message.pm, but never previously returned to the custom functions - which is all my modification does. 

The `IsSpam()` function is called irrespective of MCP as far as I can see - I haven't looked in great detail though.

PR #666, opened 2017-03-21
Correct session variable name in filters

shawniverson:
Fixes issue with adding filters in user_manager.php


shawniverson:
Bad PR, disregard.

PR #670, opened 2017-03-21
Fix for audit_log

asuweb:
Fixes audit log - logic was the wrong way round


stefaweb:
It was more fun like that!

`if (false === checkToken($_POST['token'])) { die('No! Bad dog no treat for you!'); }`

stefaweb:
conflict!

asuweb:
> conflict!

Fixed now

PR #679, opened 2017-03-22
Fix comment too long in upgrade.php

stefaweb:


upgrade.php
```diff
@@ -438,7 +438,7 @@ function color($string, $color = '')
     }
 
     // Add new column and index to maillog table
-    echo pad(' - Add maillog_id field, rblspamreport and primary key to `maillog` table');
+    echo pad(' - Add maillog_id field, rblspamreport, primary key to `maillog` table');
```
endelwar: rblspamreport has to be removed from the comment, it's not checked in this step but in the one below

Issue #664, opened 2017-03-21
Viewmail from quarantine reports


shawniverson:
Opening this issue to track this.  In the quarantine reports, a URL to viewmail.php tries to pass the ID, and will fail because a session token does not exist.

To get this resolved short term (quickly).  I am preparing a PR to disable the token check on viewmail.php.

However, this issue should remain open and another PR be prepared soon because there should be a mechanism similar to the auto release token to also view mail in a fashion that is more secure.

shawniverson:
Continuing work on this now.

Hope to have a PR ready for a permanent solution today.

PR #682, opened 2017-03-22
viewmail.php token for quarantine reports

shawniverson:
Fix for Issue #664

### DIFF
--- a/MailScanner_perl_scripts/MailWatch.pm
+++ b/MailScanner_perl_scripts/MailWatch.pm
@@ -41,6 +41,7 @@ use Storable(qw[freeze thaw]);
 use POSIX;
 use Socket;
 use Encoding::FixLatin qw(fix_latin);
+use Digest::SHA1;
 
 # Uncommet the folloging line when debugging MailWatch.pm
 #use Data::Dumper;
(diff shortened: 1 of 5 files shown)


MailScanner_perl_scripts/MailWatch.pm
```diff
@@ -345,7 +347,13 @@ sub MailWatchLogging {
     my ($todomain, @todomain);
     @todomain = @{$message->{todomain}};
     $todomain = $todomain[0];
-
+    
+    # Generate token for mail viewing
+    my ($token, $sha1);
+    $sha1 = Digest::SHA1->new;
+    $sha1->add($message->{id}, $timestamp, $message->{size});
```
endelwar: what about adding `headers` to the hash for more uniqueness?

shawniverson: Good idea, adding.

create.sql
```diff
@@ -125,6 +125,7 @@ CREATE TABLE IF NOT EXISTS `maillog` (
   `headers` mediumtext COLLATE utf8_unicode_ci,
   `quarantined` tinyint(1) DEFAULT '0',
   `rblspamreport` mediumtext COLLATE utf8_unicode_ci DEFAULT NULL,
+  `token` mediumtext COLLATE utf8_unicode_ci DEFAULT NULL,
```
endelwar: `CHAR(40)` (or `CHAR(64)` for future sha256 support) should be better than mediumtext

shawniverson: yeah, CHAR(64) sounds like a good idea

PR #690, opened 2017-03-23
Cleanup of orphaned user filters

### LINKED ISSUE #689: Function to clean filter in upgrade.php
In one of the issue I reported last week or ten days ago, we discovered that domain(s) was not deleted from filter when we delete an admin-domain or admin.

If someone can handle this and add a function in upgrade.php to clean orphan domain from the table.

shawniverson:
Fixes #689

### DIFF
--- a/upgrade.php
+++ b/upgrade.php
@@ -418,6 +418,12 @@ if ($link) {
     executeQuery($sql);
 
     echo PHP_EOL;
+    
+    // Cleanup orphaned user_filters
+    echo pad(' - Cleanup orphaned user_filters');
+    $sql = 'DELETE FROM `user_filters` WHERE `username` NOT IN (SELECT `username` FROM `users`)';
+    executeQuery($sql);
+    echo color(' DONE', 'lightgreen') . PHP_EOL;
 
     // Add new column and index to audit_log table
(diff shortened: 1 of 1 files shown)


upgrade.php
```diff
@@ -418,6 +418,24 @@ function color($string, $color = '')
...
+    $sql = 'SELECT `id`,`username` from `user_filters`';
+    $result = dbquery($sql);
+    $rows = $result->num_rows;
+    if ($rows > 0) {
+        while ($row = $result->fetch_object()) {
+            $sql = 'SELECT `username` FROM `users` WHERE `username`=\'' . trim($row->username) . '\'';
+            $result2 = dbquery($sql);
+            $rows = $result2->num_rows;
+            if ($rows === 0) {
+                $sql='DELETE FROM `user_filters` WHERE `id`=\'' . trim($row->id) . '\'';
```
Skywalker-11: How about something like this (not fully tested):
```
DELETE FROM `user_filters` WHERE `username` NOT IN (SELECT `username` FROM `users`)
```

shawniverson:
It appears to work fine on my setup.  All my orphaned user filters are gone now.

PR #701, opened 2017-03-24
Fixes for rep_audit_log.php

shawniverson:
Missing WHERE clause in SQL query
Some refactoring
Loosen validation on action field


mailscanner/rep_audit_log.php
```diff
@@ -58,6 +58,10 @@
     }
 
     $auditFilter = '';
+    $startDate='';
+    $endDate='';
+    $ipaddress = '';
+    $actions ='';
```
Skywalker-11: username was missing. done in #704

PR #712, opened 2017-03-28
Fixes #711

### LINKED ISSUE #711: Bugs in user_manager.php
The following bugs exist - I'm working on fixes:

1) Domain admins can add their own filters - this is a security issues, as they could elevate their own privileges to see other domains.  Domain admins should not be allowed to amend their own filters.

2) Domain admins should be allowed to create new users for both their own domain and domains that they have in their filters.   For example, a domain admin for `test.com` who has a filter for domain `test1.com` should be allowed to add users for `test1.com` as well as `test.com`  - This could be made optional via the config if people prefer?

asuweb:
Prevents domain admins from editing their own filters
Allows domain admins to view,edit,add,delete users in their filter domains

### DIFF
--- a/mailscanner/user_manager.php
+++ b/mailscanner/user_manager.php
@@ -565,13 +565,14 @@ if ($_SESSION['user_type'] === 'A' || $_SESSION['user_type'] === 'D') {
                     while ($row = $result->fetch_object()) {
                         echo ' <TR><TD>' . $row->filter . '</TD><TD>' . $row->active . '</TD> ';
                         if ($_SESSION['user_type'] === 'D' && $id === $_SESSION['myusername']) {
-                            echo '</TR>' . "\n";
+                            echo '<TD>' . __('nofilteraction12') . '</TD></TR>' . "\n";
                         } else {
                             echo '<TD>' . $row->actions . '</TD></TR>' . "\n";
                         }
                     }
                 }
-                if ($_SESSION['user_type'] !== 'D' || ($_SESSION['user_type'] === 'D' && $id !== $_SESSION['myusername'])) {
+                // Prevent domain admins from altering their own filters
+                if ($_SESSION['user_type'] === 'A' || ($_SESSION['user_type'] === 'D' && $id !== $_SESSION['myusername'])) {
                     echo ' <TR><TD><INPUT TYPE="text" NAME="filter"></TD><TD><SELECT NAME="active"><OPTION VALUE="Y">' . __('yes12') . '<OPTION VALUE="N">' . __('no12') . '</SELECT></TD><TD><INPUT TYPE="submit" VALUE="' . __('add12') . '"></TD></TR>' . "\n";
                 }
                 echo '</TABLE><BR>' . "\n";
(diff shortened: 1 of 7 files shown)


mailscanner/user_manager.php
```diff
@@ -536,10 +563,17 @@ function change_state(id, filter) {
...
+                        echo ' <TR><TD>' . $row->filter . '</TD><TD>' . $row->active . '</TD> ';
+                        if ($_SESSION['user_type'] === 'D' && $id === $_SESSION['myusername']) {
+                            echo '</TR>' . "\n";
+                        } else {
+                            echo '<TD>' . $row->actions . '</TD></TR>' . "\n";
+                        }
                     }
                 }
-                echo ' <TR><TD><INPUT TYPE="text" NAME="filter"></TD><TD><SELECT NAME="active"><OPTION VALUE="Y">' . __('yes12') . '<OPTION VALUE="N">' . __('no12') . '</SELECT></TD><TD><INPUT TYPE="submit" VALUE="' . __('add12') . '"></TD></TR>' . "\n";
+                if ($_SESSION['user_type'] !== 'D' || ($_SESSION['user_type'] === 'D' && $id !== $_SESSION['myusername'])) {
```
Skywalker-11: I suggest to use `$_SESSION['user_type'] === 'A'` instead of `$_SESSION['user_type'] !== 'D'` for better understanding of the code and to print a message if  a domain admin tries to edit his own filter that he is not allowed to do

Skywalker-11:
Line 545-549 should also checke if domain admin tries to edit his own filter

mailscanner/user_manager.php
asuweb: Yes, ok, I agree

asuweb:
> Line 545-549 should also check if domain admin tries to edit his own filter
The link for editing the filter won't be present, so I can't see how this case could happen?

stefaweb:
For fr.php

'nofilteraction12' => 'Action non autorisée,

Skywalker-11:
He just have to edit a filter of one of his users and then change the hidden id input field to his own id

asuweb:
@Skywalker-11 - yes, this is true. The same can be done for admins.

The same vulnerability exists in the case of editing a user - and probably also exists anywhere else we use hidden fields.  

I'm not a fan of hidden fields that are used to validate or control the actions of a form submission.  We should store these in the session, which would prevent malicious alterations. from occuring

Skywalker-11:
for de: `Sie verfügen nicht über die notwendig Berechtigung`

Also the formcheck should be used on user delete and for every action with the filters

shawniverson:
Prior to setting everything to POST, these were all using the GET method, and all of these inputs were easily manipulated via the very visible URL bar 😢 

Storing some of this in the session instead is definitely needed.

In order to use the form check everywhere here, several elements are going to need to be converted from links to actual form elements.

asuweb:
@shawniverson - I agree that hidden fields are better than the more visible GET urls, but that's "security by obscurity" rather than a secure solution.

It's going to take a lot of code re-writes to actually make it secure, but it will probably be time well spent in the long run.

Unless @endelwar would rather live with it for now, and fix it fully in v2?

endelwar:
> Unless @endelwar would rather live with it for now, and fix it fully in v2?

this is a security issued that must be resolved in 1.2.x, as usual the rule is to not trust user input at all, even validated one: so after validating input we should check that the user can work only on the field he has rights on and that the content comply with its permissions

asuweb:
@endelwar - yes, I agree it needs fixed in 1.2.x.  Storing the hidden fields in the session would be the best / quickest option of fixing it in my opinion.

PR #729, opened 2017-04-04
fix 726

Skywalker-11:


spec1re:
I can confirm this is working, thanks! :)

PR #734, opened 2017-04-05
Fixes session conflict

### LINKED ISSUE #730: Use a diffrent type of session cookie?
I have a little annoying problem here, I run a few different services on the same box where MailWatch runs, so every time I log out of MailWatch I also get log out of my other Web interface and vice versa There must be some kind of php session conflict. 

Perhaps this could be a source for many other problems, which relay on php session/tokens and so on?
```

FreeBSD 10.3-RELEASE-p17
MailWatch Version 1.2.0
MailScanner Version 4.85.2
ClamAV Version 0.99.2
SpamAssassin Version 3.4.1
PHP Version 5.6.30
MySQL Version 10.1.21-MariaDB
NGINX 1.10.2
```

asuweb:
Fixes session conflict by setting `session_name`.  It has been reported on php.net that `session_name` can cause a performance hit, so I have made this optional in `conf.php` - see comments in `conf.php.example`

Fixes #730

### DIFF
--- a/mailscanner/conf.php.example
+++ b/mailscanner/conf.php.example
@@ -39,6 +39,11 @@ define('DEBUG', false);
 // Define language (de, en, fr, it, nl, pt_br);
 define('LANG', 'en');
 
+//Session Handling - conflicts can exist when the your environment makes use of multiple php sessions on the same server
+// to resolve this, uncomment the following option.  See https://github.com/mailwatch/MailWatch/issues/730 for more info
+//define('SESSION_NAME', 'MailWatch');
+
+
 // Database settings
 //
 // As this file might be publically readable. It might be very userful to
--- a/mailscanner/lib/pear/Pager/Common.php
+++ b/mailscanner/lib/pear/Pager/Common.php
@@ -1627,6 +1627,9 @@ abstract class Pager_Common
         $this->_perPage = max($this->_perPage, 1); //avoid possible user errors
 
         if ($this->_useSessions && !isset($_SESSION)) {
+            if (defined('SESSION_NAME')) {
+                session_name(SESSION_NAME);
+            }
             session_start();
         }
         if (!empty($_REQUEST[$this->_sessionVar])) {
--- a/mailscanner/lib/pear/Pager/Common.php
+++ b/mailscanner/lib/pear/Pager/Common.php
@@ -1627,10 +1627,7 @@ abstract class Pager_Common
         $this->_perPage = max($this->_perPage, 1); //avoid possible user errors
 
         if ($this->_useSessions && !isset($_SESSION)) {
-            if (defined('SESSION_NAME')) {
-                session_name(SESSION_NAME);
-            }
-            session_start();
+           // session_start();
         }
         if (!empty($_REQUEST[$this->_sessionVar])) {
             $this->_perPage = max(1, (int)$_REQUEST[$this->_sessionVar]);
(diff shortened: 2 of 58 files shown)


stefaweb:
No SESSION_NAME in  `$optional = array()` of functions.php?

asuweb:
Added in last commit

Skywalker-11:
Is there a reason why to add it on every page? For most files it should be enough if you put it in functions.php

asuweb:
@Skywalker-11 - as per the php manual, it needs to be called directly before every call to `session_start()` which is called on nearly every page.

Skywalker-11:
@asuweb you are right that it has to be called before session_start(). But the functions.php is included before the session_start() so it should be fine. 

Btw. @shawniverson is there a reason the session_start() also is not in functions.php?

shawniverson:
No, there is not that I can recall.  It can probably be and probably should
be refactored, I would think.
On Apr 5, 2017 11:03 AM, "Michael Pietsch" <notifications@github.com> wrote:

@asuweb <https://github.com/asuweb> you are right that it has to be called
before session_start(). But the functions.php is included before the
session_start() so it should be fine.

Btw. @shawniverson <https://github.com/shawniverson> is there a reason the
session_start() also is not in functions.php?

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/mailwatch/MailWatch/pull/734#issuecomment-291909976>,
or mute the thread

asuweb:
@shawniverson @Skywalker-11 

I'll refactor it so that session_start is only called in functions.php and not on every page, and see that it still works.

asuweb:
I've refactored this

mailscanner/conf.php.example
```diff
@@ -39,6 +39,11 @@ define('DEBUG', false);
 // Define language (de, en, fr, it, nl, pt_br);
 define('LANG', 'en');
 
+//Session Handling - conflicts can exist when the your environment makes use of multiple php sessions on the same server
+// to resolve this, uncomment the following option.  See https://github.com/mailwatch/MailWatch/issues/730 for more info
+//define('SESSION_NAME', 'MailWatch');
+
+
```
stefaweb: Double blank line to remove.

`// Session Handling` missing space after //

Not sure, it really needed to ref to #730.

mailscanner/lib/pear/Pager/Common.php
```diff
@@ -1627,7 +1627,7 @@ public function setOptions($options)
         $this->_perPage = max($this->_perPage, 1); //avoid possible user errors
 
         if ($this->_useSessions && !isset($_SESSION)) {
-            session_start();
+           // session_start();
```
stefaweb: Delete?

asuweb: I left it in but commented as it's a pear package, just incase we import an update version of the package.

There should be no case where that condition will ever occur as the session is started long before we use the pear package.

mailscanner/conf.php.example
asuweb: I just thought I'd reference it as it possibly better explains when you might need to use it, and also why we've done it in the first place, but happy to remove if wished.

stefaweb: ok

mailscanner/lib/pear/Pager/Common.php
stefaweb: ok

Issue #722, opened 2017-03-31
SA Learning and Releasing not working in detail.php


saxos1983:
When I try to SA learn or release a message, nothing happens when I press the Submit button.
The same page is shown but no information text if the action was successful.
The message is also not released.

The actions are working in rep_message_ops.php under Reports --> Message Actions

This worked before the upgrade to 3.0.1.9.

shawniverson:
Please post this over at https://efa-project.org since it appears to be
related to eFa.

spec1re:
I have the same issue here with MailWatch and FreeBSD.

shawniverson:
In that case, I'll pull latest develop and see what is happening...

shawniverson:
I am having trouble reproducing this issue on latest develop:

Release: message released to shawniverson@example.org
SA Learn Learned tokens from 1 message(s) (1 message(s) examined)

shawniverson:
Can someone share a message header (with sensitive info removed) of a message that will not release in detail.php?

shawniverson:
Oh, I was on the wrong page!  Yeah, found the issue.

shawniverson:
Looks like ID isn't getting passed to do_message_ops.php properly....troubleshooting.

shawniverson:
Problem found, preparing PR

spec1re:
SA Learning still broken.

stefaweb:
@spec1re 

Could you check with 1.2.2 version?

spec1re:
@stefaweb 

sure give me a few mins.

spec1re:
`Cannot open MailScanner configuration file (dienomsconf03)`

:(

spec1re:
```
php /usr/local/www/mailscanner/upgrade.php
Cannot open MailScanner configuration file (dienomsconf03)
```
Same for upgrade.php

Skywalker-11:
Where is your MailScanner config file located?
Can you check if `MS_CONFIG_DIR` in conf.php of MailWatch points to the correct folder?

spec1re:
```
/usr/local/etc/MailScanner/MailScanner.conf

define('MS_CONFIG_DIR', '/usr/local/etc/MailScanner/');
```

Skywalker-11:
What are the permissions of the MailScanner config and are they readable to the user you are running the upgrade script and the webserver?
`ls -l /usr/local/etc/MailScanner/MailScanner.conf`
[...]

PR #735, opened 2017-04-05
fix sa-learning in languages containing special chars in submit text …

Skywalker-11:
…(see #722)


spec1re:
This is working, thanks,

PR #744, opened 2017-04-11
Change session timeout technique

### LINKED ISSUE #730: Use a diffrent type of session cookie?

shawniverson:
Fixes #730

### DIFF
--- a/mailscanner/functions.php
+++ b/mailscanner/functions.php
@@ -100,18 +100,14 @@ ini_set('session.cookie_httponly', 1);
 ini_set('session.use_only_cookies', 1);
 ini_set('session.use_trans_sid', 0);
 
-// Session garbage collection 5 minutes based on user activity
-// or STATUS_REFRESH + 60 sec, whichever is greater.
+// Session timeout 10 minutes based on user activity
+// or STATUS_REFRESH + 300 sec, whichever is greater.
 
-if (defined(STATUS_REFRESH) && STATUS_REFRESH + 60 > 300) {
-    ini_set('session.gc_maxlifetime', STATUS_REFRESH + 60);
-    define('SESSION_TIMEOUT', STATUS_REFRESH + 59);
+if (defined(STATUS_REFRESH) && STATUS_REFRESH + 300 > 600) {
+    define('SESSION_TIMEOUT', STATUS_REFRESH + 300);
 } else {
-    ini_set('session.gc_maxlifetime', 300);
-    define('SESSION_TIMEOUT', 299);
+    define('SESSION_TIMEOUT', 600);
 }
-ini_set('session.gc_divisor', 1);
-ini_set('session.gc_probability', 1);
 
 $session_cookie_secure = false;
 if (SSL_ONLY === true) {
@@ -127,6 +123,7 @@ if (defined('SESSION_NAME')) {
 session_set_cookie_params(0, $params['path'], $params['domain'], $session_cookie_secure, true);
 unset($session_cookie_secure);
 session_start();
+
 // set default timezone
 date_default_timezone_set(TIME_ZONE);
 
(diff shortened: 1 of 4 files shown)


shawniverson:
Hold on merge, testing.

mailscanner/logout.php
```diff
@@ -51,4 +51,8 @@
 
 session_destroy();
 
-header('Location: index.php');
+if (isset($_GET['error'])) {
+    header('Location: login.php?error=' . $_GET['error']);
```
endelwar: filter this $_GET fox xss

mailscanner/functions.php
```diff
@@ -287,6 +286,12 @@ function html_start($title, $refresh = 0, $cacheable = true, $report = false)
         }
     }
 
+    if (!isset($_SESSION['last_update']) || $_SESSION['last_update'] + SESSION_TIMEOUT < time()) {
```
Skywalker-11: This would always be true because the last_update has already been overwritten in line 127

shawniverson:
Yeah, I have to rethink this logic...
On Apr 11, 2017 6:23 AM, "Michael Pietsch" <notifications@github.com> wrote:

*@Skywalker-11* commented on this pull request.
------------------------------

In mailscanner/functions.php
<https://github.com/mailwatch/MailWatch/pull/744#discussion_r110866204>:

> @@ -287,6 +286,12 @@ function html_start($title, $refresh = 0, $cacheable = true, $report = false)
         }
     }

+    if (!isset($_SESSION['last_update']) || $_SESSION['last_update']
+ SESSION_TIMEOUT < time()) {

This would always be true because the last_update has already been
overwritten in line 127

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/mailwatch/MailWatch/pull/744#pullrequestreview-32066821>,
or mute the thread

mailscanner/functions.php
shawniverson: Moved line 127 to checklogin.php

shawniverson:
Ready to test some more...

shawniverson:
Adjusting values to test for a timing issue...

stefaweb:
StyleCI

shawniverson:
@stefaweb I don't agree with StyleCl's anysis here.  It is getting thrown off by the HTML/PHP mix, I think...

shawniverson:
@endelwar After testing this with the help of @spec1re, we are seeing inconsistent results at 5 minutes for the session timeout.  However, at 10 minutes, we are both able to get consistent results.  Seems browser related.  The PR is ready for your review.

endelwar:
Looks good to me, I agree on StyleCI gone crazy. Merging right now.

Skywalker-11:
Is there a way to exclude the file from styleci?

endelwar:
@Skywalker-11 I've just written to the StyleCI developer asking for correctness of the analisys

Skywalker-11:
@endelwar any news on this?

endelwar:
I'm reporting here the reply, but I don't agree with the "view" issue:
```
Hi Manuel,

Thanks for getting in touch. StyleCI is not designed to fix "view" files. It treats them as source code, so, the indentation change there is correct, it just looks funny, because it's in a view file.

I'd recommend excluding your view files from analyses.
```

We should refactor login.php, maybe by indenting better the php code part or convert html to php strings.

PR #750, opened 2017-04-13
Adds the ability for users to select the language of the gui

Skywalker-11:
Can give the users the ability to change the gui language after login.
Feature will be enabled if USER_SELECTABLE_LANG is defined and more than one language code is given.

Language codes are the names of the php files under mailscanner/languages/

On most pages the language change should work without any problems. Only pages that uses POST to submit security tokens will show the error message of an invalid token.


stefaweb:
What is happening with audit Log ?
Do the language of the logged messages will be the one of the admin or the one of user?
We may be need to define a politic for this!

Skywalker-11:
I haven't thaught of that but with the current version it would be logged with the user selected language.

We have to load multiple languages if the user has selected not the default one.

stefaweb:
We can may be change the way to manage this?

- Instead of storing a clear message in database, we only store a message number.
- We print the right language message in GUI according to the corresponding number and the selected language.

Skywalker-11:
Problem is that in some cases we store content related things like which user created a password reset.

I thought on using a second global variable $audit_lang or $system_lang that will be used for the audit stuff which contains the language from LANG in the conf.php (not necessary the user selected lang) and to add an additional parameter to the function `__($string, $audit_log=false) {...} ` which will use that $audit_lang if $audit_log is true.

stefaweb:
Not easy all this!
We can have the MailWatch admin with a language, but also a domain-admin with another language, and a user with another language... We can also have a user without domain-admin...

Skywalker-11:
That would not be a problem because domain admins and normal users are not able to view the audit log and you can assume that the server admins speak the same language that can be used for all audit_logs which will be set by LANG in conf.php.

The gui language can be changed by every user for each browser separately (stored with cookie). It doesn't matter what type of user it is.

Language selection on login page is still missing atm. Any suggestion where to place it?

PR #755, opened 2017-04-18
Remove comma line 4219

stefaweb:


stefaweb:
Error. Comma seems now needed.

PR #769, opened 2017-04-28
Fixes Path to Functions

asuweb:
Fixes path to functions to adhere to standard path

### DIFF
--- a/tools/Cron_jobs/mailwatch_db_clean.php
+++ b/tools/Cron_jobs/mailwatch_db_clean.php
@@ -31,7 +31,7 @@
  */
 
 // Edit if you changed webapp directory from default
-$pathToFunctions = '/opt/mailwatch/mailscanner/functions.php';
+$pathToFunctions = '/var/www/html/mailscanner/functions.php';
 if (!@is_file($pathToFunctions)) {
     die('Error: Cannot find functions.php file in "' . $pathToFunctions . '": edit ' . __FILE__ . ' and set the right path on line ' . (__LINE__ - 3) . PHP_EOL);
 }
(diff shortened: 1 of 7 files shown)


Skywalker-11:
Looks like @endelwar moved all paths to /opt/mailwatch: [https://github.com/mailwatch/MailWatch/commit/4eda57017b0c80e5c60cb9f9adf77553db3aea6d](https://github.com/mailwatch/MailWatch/commit/4eda57017b0c80e5c60cb9f9adf77553db3aea6d)

asuweb:
Ahh, that's going to cause a few people some issues I feel

stefaweb:
Not the best idea.

/var/www/html is well know of everybody and old mailwatch user will have to move stuff.

Skywalker-11:
@endelwar I think we should revert that commit now and apply it with 2.x or 1.3 release

endelwar:
We should correct online doc too...

stefaweb:
> I think we should revert that commit now and apply it with 2.x or 1.3 release

I agree with this. Please wait for 2.0 to change path to /opt. At present, this kind of change brings no technical advantage. It only forces everyone to spend time for nothing.

endelwar:
> this kind of change brings no technical advantage

I'm going to revert it, but I disagree about no technical advantages: having everything in a single, not web exposed, directory (apart mailscanner dir) make easier to upgrade the system, as everything is in one place and the many tools scripts can be safely symlinked to /usr/local/bin and upgraded with a single git command. Also this make easier to give support to new users, as we know where the single files are.

asuweb:
@endelwar - I agree with you on that point - but we need to work out how to handle it otherwise for many people it will break existing installations.  

Perhaps we need to incorporate it into the upgrade / install scripts

stefaweb:
I agree with you @endelwar but for 2.0 it will be be less disruptive for users.

When do you plan to start 2.0 or at minimum create the repository for it ? What is the road map for 2.0?

Issue #800, opened 2017-05-22
Problem to clean old mail with mailwatch_db_clean.php / reports and Exim


stefaweb:
With last develop.

Something seems wrong with develop version. May be the problem is there from a long time.

Running `mailwatch_db_clean.php` manually, don't purge old messages, or we have an error in reports or the database record is corrupted. 

![4c507a34-3ee7-11e7-9721-6c5634cc0c05](https://cloud.githubusercontent.com/assets/12907102/26305570/23cd4df4-3ef0-11e7-92e9-8593af47b913.png)

But I have in config.php: ` define('RECORD_DAYS_TO_KEEP', 60);`

In maillog, the oldest record look like this:

```
timestamp: 2017-03-23 20:10:54
date: 2017-01-22
time: 18:14:51
```
And in report, I got this (start with 22/01/17).
![422dd188-3ef0-11e7-87e8-2de9a8e3de27](https://cloud.githubusercontent.com/assets/12907102/26305654/7dc1d492-3ef0-11e7-8228-56475cb113c7.png)

Report print `date` but where is the true?

stefaweb:
Problem in the database.

![capture d ecran 2017-05-22 a 13 25 22](https://cloud.githubusercontent.com/assets/12907102/26306058/363cb93c-3ef2-11e7-8d7c-40ca0dd51bf6.png)

stefaweb:
The database.
If someone have an idea?

Start
![capture d ecran 2017-05-22 a 13 31 01](https://cloud.githubusercontent.com/assets/12907102/26307263/fce68f26-3ef3-11e7-9b5f-8484287c4770.png)

Change
![capture d ecran 2017-05-22 a 13 38 14](https://cloud.githubusercontent.com/assets/12907102/26307283/111ad0ce-3ef4-11e7-9d79-20c887ca2305.png)

Skywalker-11:
If you look in mailwatch_db_clean.php you can see that nothing is deleted from maillog. Only mtalog is cleared

Skywalker-11:
```
>show create table maillog;
........
CREATE TABLE `maillog` (
  `maillog_id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
  `timestamp` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
.......
```
[...]

PR #820, opened 2017-05-31
fix timestamp to not be updated on row change in maillog (#800)

Skywalker-11:


stefaweb:
And this one?

> mailwatch_db_clean.php:dbquery('DELETE LOW_PRIORITY FROM maillog WHERE timestamp < (NOW() - INTERVAL ' . RECORD_DAYS_TO_KEEP . ' DAY)');

Skywalker-11:
that is intended as it remove the old mails

PR #832, opened 2017-06-03
cleanup unused code

Skywalker-11:


mailscanner/functions.php
```diff
@@ -247,7 +247,7 @@ function mailwatch_version()
 }
 
 if (!function_exists('imageantialias')) {
-    function imageantialias($image, $enabled)
```
endelwar: sure about this? if imageantialias is not present jpgraph will still pass 2 arguments to this function

Skywalker-11: yes should work. try
```
function bla(){
 echo "bla";
}
bla("a", "b");

```

PR #837, opened 2017-06-07
Movexaxislabel

Skywalker-11:


Skywalker-11:
Just to clarify in Chart.js and Char.min.js only two `/ ` changed to `*`. Not sure why git thinks that other stuff also changed

stefaweb:
Do you plan to remove lib/jpgraph and images/cache ?

stefaweb:
> Just to clarify in Chart.js and Char.min.js only two / changed to *. Not sure why git thinks that other stuff also changed

Seems the files are full of TAB.

Skywalker-11:
> Seems the files are full of TAB.

But why does git think that the tabs changed? I don't get that  :confused:

endelwar:
> Not sure why git thinks that other stuff also changed

I think that is your ide, that (correctly :wink:) convert tabs to spaces, to see the relevant edits just add a `?w=1` at the end of compare url (https://github.com/mailwatch/MailWatch/pull/837/files?w=1)

stefaweb:
I've done a copy/paste and the soft said "Do you want to convert TAB with SPACE?".

Skywalker-11:
@endelwar are you sure? github shows that in both versions it is a mix of tabs space. Also nano usually does not change it :unamused:

endelwar:
a `set tabstospaces` in .nanorc?

Skywalker-11:
With old vi its working...

Skywalker-11:
@stefaweb yes plan is to remove jpgraph and the cache folder but rep_mcp_score_dist.php still uses this and I haven't changed that yet

PR #841, opened 2017-06-08
fix upgrade.sh to detect existing last_update field

Skywalker-11:


upgrade.php
```diff
@@ -514,6 +511,14 @@ function color($string, $color = '')
         $sql = "ALTER TABLE `maillog` ADD `released` tinyint(1) DEFAULT '0'";
         executeQuery($sql);
     }
+
+    echo pad(' - Add last_update field to `maillog` table');
+    if (check_column_exists('users', 'resetid') === false) {
```
endelwar: the check must be on `last_update`, not `resetid`

Skywalker-11: sorry was a long day

PR #843, opened 2017-06-09
Chart.js for mcp, remove JpGraph library

Skywalker-11:
- change mcp graph to Chart.js
- remove jpgraph library 
- remove cache dir and associated functions


endelwar:
I ❤️ this PR 💯 
Just 2 things left:
- delete cache dir with upgrade.php script
- delete jpGraph special exception to GPL licence in all file headers

PR #844, opened 2017-06-10
MailWatch namespace for all files (includes #843)

Skywalker-11:
Based on #843 and additionally introduce a MailWatch namespace :smile: 
Not sure if we need that now. But now that we have it...

### DIFF
--- a/mailscanner/database.php
+++ b/mailscanner/database.php
@@ -25,6 +25,7 @@
  * Software Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
  */
 
+namespace MailWatch;
 class database
 {
     /** @var mysqli $link */
@@ -45,11 +46,11 @@ class database
     {
         if (!self::$link instanceof mysqli) {
             try {
-                $driver = new mysqli_driver();
+                $driver = new \mysqli_driver();
                 $driver->report_mode = MYSQLI_REPORT_ALL;
                 set_error_handler(function () {
                 });
-                self::$link = new mysqli($host, $username, $password, $database);
+                self::$link = new \mysqli($host, $username, $password, $database);
                 restore_error_handler();
                 self::$link->options(MYSQLI_INIT_COMMAND, "SET sql_mode=(SELECT TRIM(BOTH ',' FROM REPLACE(@@sql_mode,'ONLY_FULL_GROUP_BY','')))");
                 $charset = 'utf8';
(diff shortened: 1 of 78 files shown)


tools/Cron_jobs/mailwatch_db_clean.php
```diff
@@ -28,7 +28,7 @@
 namespace MailWatch;
 
 // Edit if you changed webapp directory from default
-$pathToFunctions = '/var/www/html/mailscanner/functions.php';
+$pathToFunctions = '/opt/1.2.0/mailscanner/functions.php';
```
stefaweb: `/opt/1.2.0/mailscanner/`

???

Skywalker-11: Fixed

Skywalker-11:
By the way it is also part of the psr-1 convention for code for php < 5.3

mailscanner/database.php
```diff
@@ -41,11 +41,11 @@ private function __construct()
      * @param string $username
      * @param string $password
      * @param string $database
-     * @return mysqli
+     * @return \mysqli
      */
     public static function connect($host = '', $username = '', $password = '', $database = '')
     {
-        if (!self::$link instanceof mysqli) {
+        if (!self::$link instanceof \mysqli) {
```
stefaweb: For my knowledge, what does this `\` means?

Skywalker-11: It means that the class is in the root namespace which is the default if you are not setting a namespace in a php file. It is also the namespace where the standard php classes are. If you have a namespace set it is assumed that all classses you are using are in the mentioned namespace and if.you want to use classes in.the root namespace you have to prefix that classes with a `\`. You can also use `use ... as ...` to set a namespace for a single class in the current file

stefaweb: Thanks.

endelwar:
Well namespaces are good, but it's a bit to early ;)
I think that a modernizing refactory is what should be done for MailWatch or we will never move on to 2.0
A quick roadmap (a new post will be prepared in the next days) could be this:
- fix standing 1.2 bugs (~40 issue not tagged 2.0) ETA End of August?
- start refactoring in a different branch
- implement composer and autoloading
- consolidate all functions in separated and organized classes
- remove all globals (dependency injection) and create factories classes
- write tests!!!
- replace pear libs with something better
- make better cli script
- split domain logic from view
- implement request/response login
- reorganize directory structure (/opt/mailwatch)
- implement a router
- new UI layout

stefaweb:
Can we publish 1.2.4 before?

(When the last cosmetic changes will be over in report graphs)

asuweb:
@stefaweb - see the latest milestones https://github.com/mailwatch/MailWatch/milestones

Skywalker-11:
OK I will close this then

PR #858, opened 2017-06-14
optional config parameter to define the chart interval

Skywalker-11:


stefaweb:
Do we need to translate this in graphgenerator.inc.php?

`throw new \Exception('timeInterval or timeScale not set');`

Skywalker-11:
Exception shouldn't be thrown normally. It is just a reminder for devs when they are trying to use that function 😄

stefaweb:
Ok.  😁

Think to center top legends.

Skywalker-11:
The Legend alignment is tricky because the language strings have different lengths and this is also done by chart.js. Don't realy want to mess around with that code too much.

stefaweb:
No problem if source of problem. ;)

PR #859, opened 2017-06-15
Table for traffic graph in status.php

stefaweb:
Let me know if the wording in English is not the best.

Need DE: `    'trafficgraph03' => 'Traffic of the last %s minutes',`

capture d ecran 2017-06-15 a 18 59 02

### DIFF
--- a/mailscanner/functions.php
+++ b/mailscanner/functions.php
@@ -4689,8 +4689,8 @@ function printTrafficGraph()
      SELECT
       timestamp AS xaxis,
       1 as total_mail,
-      SUM(CASE WHEN (virusinfected+nameinfected+otherinfected) > 0 THEN 1 ELSE 0 END) AS total_virus,
-      SUM(CASE WHEN (isspam+ishighspam+issaspam+isrblspam+spamblacklisted) > 0 THEN 1 ELSE 0 END) AS total_spam,
+      virusinfected AS total_virus,
+      isspam AS total_spam,
       size AS total_size
      FROM
       maillog
--- a/mailscanner/functions.php
+++ b/mailscanner/functions.php
@@ -4689,9 +4689,8 @@ function printTrafficGraph()
      SELECT
       timestamp AS xaxis,
       1 as total_mail,
-      virusinfected AS total_virus,
-      isspam AS total_spam,
-      size AS total_size
+      sum(virusinfected+nameinfected+nameinfected) AS total_virus,
+      sum(isspam+ishighspam+issaspam+isrblspam+spamblacklisted) AS total_spam
      FROM
       maillog
      WHERE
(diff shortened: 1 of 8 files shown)


stefaweb:
This doesn't work.

```
      sum(virusinfected+nameinfected+nameinfected) AS total_virus,
      sum(isspam+ishighspam+issaspam+isrblspam+spamblacklisted) AS total_spam
```

mailscanner/functions.php
```diff
@@ -4682,9 +4689,8 @@ function printTenMinutesGraph()
      SELECT
       timestamp AS xaxis,
       1 as total_mail,
-      virusinfected AS total_virus,
-      isspam AS total_spam,
-      size AS total_size
+      sum(virusinfected+nameinfected+otherinfected) AS total_virus,
+      sum(isspam+ishighspam+issaspam+isrblspam+spamblacklisted) AS total_spam
```
endelwar: I don't think that this sum will do what you want: if you look at [mailscanner code](https://github.com/MailScanner/v5/blob/b02c2fbbfd7131a2b7803f72c0129b2f46c7255e/common/usr/share/MailScanner/perl/MailScanner/Message.pm) isspam is always set to 1 in any case the massage is spam (low, high rbl, etc...)

stefaweb: I was adding cabbages and carrots ...

stefaweb: Do you have an idea?

endelwar: you have to revert total_spam as before your edit, and you can use a `CASE` in total_virus:
```sql
SUM(CASE WHEN (virusinfected+nameinfected+otherinfected) > 0 THEN 1 ELSE 0 END) AS total_virus
```

stefaweb: Seems to work but I've not enough records to test.
Thanks.

stefaweb: In fact doesn't work. New traffic do not generated new pic in the graph.

stefaweb:
In fact new query do not work. New mails are not printed at all.

```
SUM(CASE WHEN (virusinfected+nameinfected+otherinfected) > 0 THEN 1 ELSE 0 END) AS total_virus,
SUM(CASE WHEN (isspam+ishighspam+issaspam+isrblspam+spamblacklisted) > 0 THEN 1 ELSE 0 END) AS total_spam,
```

Skywalker-11:
@stefaweb time interval over an hour would result in the same index which is why >60 min won't work. Changing the format to `H:i` could work (not tested yet)

stefaweb:
May be @endelwar can merge my "cosmetic" PR and you takes over the stuff.

Skywalker-11:
de:  'trafficgraph03' => 'Traffic der letzten %s Minuten',

PR #862, opened 2017-06-16
use HH:mm format for labels; fixes scale limited to 60min

Skywalker-11:


stefaweb:
Seems ok for 120 and 240 minutes.
We still have a gap with 60 minutes.

Can be possible to reduce space between the time legend and the bottom frame?

capture d ecran 2017-06-16 a 09 54 00
capture d ecran 2017-06-16 a 09 53 40
capture d ecran 2017-06-16 a 09 53 24

PR #863, opened 2017-06-16
move code for traffic graph to correct function; fix header if STATUS…

Skywalker-11:
…GRAPH_INTERVAL is undefined, reduce graph spacing


stefaweb:
My PR #864 crossed yours.

If you like the idea, take it and I'll delete my PR.

stefaweb:
Does `size AS total_size` needed?

> $graphgenerator->sqlQuery = '
>        SELECT
>        timestamp AS xaxis,
>        1 as total_mail,
>        virusinfected AS total_virus,
>        isspam AS total_spam,
>        size AS total_size
>       FROM
>        maillog
>       WHERE
>        1=1
>       AND
>        timestamp BETWEEN (NOW() - INTERVAL ' . $graphInterval . ' MINUTE) AND NOW()
>       ORDER BY
>        timestamp DESC
>      ';

mailscanner/conf.php.example
```diff
@@ -301,5 +301,5 @@ define('RESET_LINK_EXPIRE', 1);
 // define('PWD_RESET_FROM_ADDRESS', 'support@yourdomain.com'); // Set is you wish reset email sent from alternative address.
 /* End Password Reset */
 
-//Interval for the traffic chart for admins (in minutes) default is 60
+// Interval for the traffic chart for admins (in mulitple of 60 minutes) default is 60
```
stefaweb: multiple

stefaweb:
What do think of this?

```
     SELECT
      timestamp AS xaxis,
      1 as total_mail,
      CASE 
      WHEN virusinfected > 0 THEN 1
      WHEN nameinfected > 0 THEN 1
      WHEN otherinfected > 0 THEN 1
      ELSE 0 END AS total_virus,
      CASE 
      WHEN isspam > 0 THEN 1
      WHEN ishighspam > 0 THEN 1
      WHEN issaspam > 0 THEN 1
      WHEN isrblspam > 0 THEN 1
      WHEN spamblacklisted > 0 THEN 1 
      ELSE 0 END AS total_spam
     FROM
      maillog
     WHERE
      1=1
     AND
      timestamp BETWEEN (NOW() - INTERVAL 120 MINUTE) AND NOW()
     ORDER BY
      timestamp DESC
```

mailscanner/style.css
```diff
@@ -932,4 +932,6 @@ select.lang > option {
 #trafficgraph {
     width: auto;
     height: 180px;
+    margin-top: -4px;
+    margin-bottom: -13px;
```
stefaweb: Add `margin-right: -4px;`

Skywalker-11:
This looks good
```
      WHEN virusinfected > 0 THEN 1
      WHEN nameinfected > 0 THEN 1
      WHEN otherinfected > 0 THEN 1
      ELSE 0 END AS total_virus,
``` 

But I think that this is unnecessary because as soon as they are in one of them the `isspam` is always true
```
      WHEN isspam > 0 THEN 1
      WHEN ishighspam > 0 THEN 1
      WHEN issaspam > 0 THEN 1
      WHEN isrblspam > 0 THEN 1
      WHEN spamblacklisted > 0 THEN 1 
      ELSE 0 END AS total_spam
```

mailscanner/style.css
Skywalker-11: done

stefaweb:
Ok for Query.

Question, is normal to have some columns NULL (not 0)?

capture d ecran 2017-06-16 a 17 24 10

I have not checked all concerned columns.

Skywalker-11:
yes it is normal. At least it's the same here

stefaweb:
Possible to have this kind of time legend?

legend

Skywalker-11:
Theoretically yes but its not trivial.

stefaweb:
And use the curent code but only print one value out of two?

Skywalker-11:
The problem is that I only define what labels are there  but which label will be printed and how is Chart.js intern

stefaweb:
Try this.

https://stackoverflow.com/questions/22064577/limit-labels-number-on-chart-js-line-chart

stefaweb:
If you use  `ticks: { autoSkip: true, maxTicksLimit: 8 },` for xAxes, you got:

capture d ecran 2017-06-17 a 08 38 30

To evenly distribute ticks:

https://stackoverflow.com/questions/37250456/chart-js-evenly-distribute-ticks-when-using-maxtickslimit/37257056#37257056

Skywalker-11:
thanks i will take a look at that

stefaweb:
Need new style.

```
#trafficgraph {
    width: auto;
    height: 180px;
    margin-top: -4px;
    margin-right: -4px;
}
```

Skywalker-11:
I changed the line graphs to use the time scales which fixes the gap in the labels. 
Still have the problem that the latest interval seems empty because the data point is aligned to the left of it. Going to look at it next week

Issue #848, opened 2017-06-12
Traffic graph in status.php


stefaweb:
One day, may be?   😎

A real time graph with the minimum (mail trafic, total & clean for the running day).

![capture d ecran 2017-06-12 a 17 18 21](https://user-images.githubusercontent.com/12907102/27041722-073e1dc0-4f95-11e7-8eec-fb2dd82bba37.png)

Skywalker-11:
Give me the sql query and I give you the graph 😀

stefaweb:
We can may be use this one?

![capture d ecran 2017-06-12 a 17 49 08](https://user-images.githubusercontent.com/12907102/27042840-6ae1bd3e-4f98-11e7-9aaa-5b9f99d60856.png)

Skywalker-11:
should it be line or bar graphs?

stefaweb:
Seems better with lines.

An example from Proxmox.

![capture d ecran 2017-06-13 a 07 51 40](https://user-images.githubusercontent.com/12907102/27067868-5dc1edd4-500d-11e7-8f49-64ac4df63399.png)

In Proxmox, we can select hour, day, week, month, year.

stefaweb:
Great job @Skywalker-11. 

Try to arrange the layout.
Try to add a frame box with a yellow title around the graph.

May be to find space, we can change the way the colors box is printed in GUI. We can may be have the color on one line as for the report legends.

![capture d ecran 2017-06-14 a 08 16 11](https://user-images.githubusercontent.com/12907102/27118311-f8b98158-50da-11e7-99a7-a7988000e8e4.png)
![capture d ecran 2017-06-14 a 08 15 51](https://user-images.githubusercontent.com/12907102/27118312-f8dfa590-50da-11e7-8e45-948ef003c16a.png)

stefaweb:
More on this.

As I've not enough trafic on my dev server, I'm not sure to understand what is printing the graph. Can you explain?

stefaweb:
Sometime, I got this:

```
Notice: Undefined offset: 42 in /var/www/html/mailscanner/graphgenerator.inc.php on line 361
Notice: Undefined offset: 42 in /var/www/html/mailscanner/graphgenerator.inc.php on line 361
Notice: Undefined offset: 42 in /var/www/html/mailscanner/graphgenerator.inc.php on line 361
```

Skywalker-11:
Graph prints the total mail, spam and virii counts for the last ten minutes. 
[...]

PR #873, opened 2017-06-24
possibility to detect mysql or mariadb in upgrade.php

Skywalker-11:
fix for schema update not working correctly mentioned in #848

### DIFF
--- a/upgrade.php
+++ b/upgrade.php
@@ -242,10 +242,9 @@ function getSqlServer()
 {
     global $link;
     //test if mysql or mariadb is used.
-    $sql = 'SHOW variables WHERE variable_name="aria_block_size"';
+    $sql = 'SELECT VERSION() as version"';
     $result = $link->query($sql);
-    $result_count = $result->num_rows;
-    if ($result_count === 0) {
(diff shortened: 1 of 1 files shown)


stefaweb:
This is may be better to easy find engine/version of server?

```
> SELECT VERSION() as mysql_version;
+--------------------------+
| mysql_version            |
+--------------------------+
| 10.1.24-MariaDB-1~jessie |
+--------------------------+
```

```
mysql> SELECT VERSION() as mysql_version;
+-----------------+
| mysql_version   |
+-----------------+
| 5.5.55-0+deb7u1 |
+-----------------+
1 row in set (0.00 sec)
```

upgrade.php
```diff
@@ -242,10 +242,9 @@ function getSqlServer()
 {
     global $link;
     //test if mysql or mariadb is used.
-    $sql = 'SHOW variables WHERE variable_name="aria_block_size"';
+    $sql = 'SELECT VERSION() as version"';
```
stefaweb: Remove ".
`$sql = 'SELECT VERSION() as version';`

Skywalker-11: done

Issue #713, opened 2017-03-29
Top virus in GUI


stefaweb:
How is working this feature?

Or this feature is really working?

I got 2 virus today but "top virus" is empty.

![capture d ecran 2017-03-29 a 17 49 33](https://cloud.githubusercontent.com/assets/12907102/24463864/4f49e69e-14a8-11e7-8d5e-e601278f3cba.png)

(with a patched RC5)

shawniverson:
I vaguely recall something up with that when more than one virus scanner
was in use....can't seem to find that thread....

Skywalker-11:
You're right only the first virus scanner is used in MailWatch

stefaweb:
I've only one virus scanner installed (clamav).

Skywalker-11:
Are they correctly shown on the report page?

stefaweb:
Yes, if I selected today and virus, I got them.

![capture d ecran 2017-03-29 a 20 55 35](https://cloud.githubusercontent.com/assets/12907102/24471359/2a38329c-14c2-11e7-952d-32b5f364ffc0.png)

stefaweb:
This was working in the past.

Skywalker-11:
Does this return something
`SELECT report FROM maillog WHERE virusinfected>0 AND date = CURRENT_DATE()`

stefaweb:
Yes.

```
Clamd: 57Callie.js was infected: Txt.Downloader.Nemucod-14748 MailScanner: JScript Scripts are dangerous in email (57Callie.js) ,Clamd: 57Callie.js was infected: Txt.Downloader.Nemucod-14748 MailScanner: JScript Scripts are dangerous in email (57Callie.js)

Clamd: 58Elfrida.js was infected: Txt.Malware.Agent-1813827 MailScanner: JScript Scripts are dangerous in email (58Elfrida.js) ,Clamd: 58Elfrida.js was infected: Txt.Malware.Agent-1813827 MailScanner: JScript Scripts are dangerous in email (58Elfrida.js)

```

endelwar:
> This was working in the past.

no, never did. the actual code return none if top 2 virus count is equal
https://github.com/mailwatch/1.2.0/blob/f79b89d1413872e0762fb88d6a5c25e300897e2b/mailscanner/functions.php#L1270-L1273

stefaweb:
Ok, I've only 2 virus. Need 3 MINIMUM...

I close.

endelwar:
[...]

PR #874, opened 2017-06-24
Support for multiple virus scanners; multiple top viruses (#713)

Skywalker-11:


mailscanner/languages/fr.php
```diff
@@ -196,6 +196,7 @@
     'barmail03' => 'E-Mails',
     'barvirus03' => 'Virus',
     'barspam03' => 'Spam',
+    'moretopviruses03' => 'and %s other viruses',
```
stefaweb: 'moretopviruses03' => 'et %s autres virus',

PR #880, opened 2017-06-24
Style Update + repair one bug

annixos:
screen


tools/Sendmail_relay/mailwatch-sendmail-relay
```diff
@@ -23,7 +23,7 @@
 # Author: @stefaweb, 14/06/2015
 
 NAME="mailwatch-sendmail-relay"
-DAEMON="/usr/local/bin/sendmail_relay.php"
+DAEMON="/usr/local/bin/mailwatch-sendmail-relay.php"
```
stefaweb: Could you place this patch in a separate PR?
This a nothing to do with color chart.

shawniverson:
Huh, I think I've seen this color scheme somewhere before.....*scratches chin*......

shawniverson:
Although I like this color scheme (it is close to or exactly like eFa), I think there is something to be said about pure MailWatch having a different color scheme.  This really does help me (and perhaps the MailWatch team) to differentiate between these types of installations instantly and visually.

tools/Sendmail_relay/mailwatch-sendmail-relay
annixos: https://github.com/mailwatch/MailWatch/pull/881

PR #883, opened 2017-06-25
Fix type mismatch in sa rule hits report

shawniverson:
Fixes issue where this if block is always true.

### DIFF
--- a/mailscanner/rep_mcp_rule_hits.php
+++ b/mailscanner/rep_mcp_rule_hits.php
@@ -21,10 +21,6 @@
  * your version of the program, but you are not obligated to do so.
  * If you do not wish to do so, delete this exception statement from your version.
  *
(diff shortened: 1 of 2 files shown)


stefaweb:
And this one?

```
rep_mcp_rule_hits.php:        if ($row->ismcp !== 0) {
```

shawniverson:
Good catch, updating PR

stefaweb:
I've found these but not sure its related.

```
msrule.php:    if ($FilePath === false || (strpos($FilePath, $MailscannerEtcDir) !== 0 && strpos($FilePath, $MailscannerRepDir) !== 0)) {
password_reset.php:            if ($result->num_rows !== 1) {
password_reset.php:                if ($result->num_rows !== 1) {
```

shawniverson:
Those are ok.  They return as numeric in those cases.

PR #885, opened 2017-06-26
option specify colors for line chart; Auto detect best ticks/grids assignment

Skywalker-11:
Also auto detectc best number of ticks and grids to create a consistent distribution of them

### DIFF
--- a/mailscanner/functions.php
+++ b/mailscanner/functions.php
@@ -4740,7 +4740,7 @@ function printTrafficGraph()
     );
     $graphgenerator->graphTitle = '';
     $graphgenerator->settings['timeInterval'] = 'PT' . $graphInterval . 'M';
-    if ($graphInterval < 240) {
+    if ($graphInterval < 360) {
         $graphgenerator->settings['timeScale'] = 'PT1M';
         $graphgenerator->settings['timeGroupFormat'] = 'Y-m-dTH:i:00';
         $graphgenerator->settings['timeFormat'] = 'H:i';
--- a/mailscanner/functions.php
+++ b/mailscanner/functions.php
@@ -4721,13 +4721,13 @@ function printTrafficGraph()
     $graphgenerator->graphColumns = array(
         'labelColumn' => 'time',
         'dataLabels' => array(
-            array(__('barmail03'), __('barvirus03'), __('barspam03')),
+            array(__('barvirus03'), __('barspam03'), __('barmail03')),
         ),
         'dataNumericColumns' => array(
-            array('total_mailconv', 'total_virusconv', 'total_spamconv'),
+            array('total_virusconv', 'total_spamconv', 'total_mailconv'),
         ),
         'dataFormattedColumns' => array(
-            array('total_mailconv', 'total_virusconv', 'total_spamconv'),
+            array('total_virusconv', 'total_spamconv', 'total_mailconv'),
         ),
         'xAxeDescription' => '',
         'yAxeDescriptions' => array(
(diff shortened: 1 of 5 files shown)


mailscanner/lib/lineConfig.js
```diff
@@ -1,14 +1,32 @@
...
+var spamColor   = '#EE6262'; // red
+var volumeColor = '#f5d932'; // yellow
+/*******************************/
+
+var defaultColors= [
+  mailColor,
+  virusColor,
+  spamColor,
+  volumeColor,
   '#b9e3f9' // light blue
```
stefaweb: Do we need this one?

Skywalker-11: Was for mcp. Applied that for total mail report

stefaweb:
To work with last lineConfig.js code. I reduced a little bottom margin and adjusted left.

style.css
```
#trafficgraph {
    width: auto;
    height: 180px;
    margin-top: -4px;
    margin-bottom: -12px;
    margin-right: -2px;
    margin-left: 3px;
}
```

mailscanner/functions.php
```diff
@@ -4740,7 +4740,7 @@ function printTrafficGraph()
     );
     $graphgenerator->graphTitle = '';
     $graphgenerator->settings['timeInterval'] = 'PT' . $graphInterval . 'M';
-    if ($graphInterval < 240) {
+    if ($graphInterval < 360) {
```
stefaweb: Hello @Skywalker-11  

Why do you stay with the test?

I will remove it completely.

With 
```
        $graphgenerator->settings['timeScale'] = 'PT1M';
        $graphgenerator->settings['timeGroupFormat'] = 'Y-m-dTH:i:00';
        $graphgenerator->settings['timeFormat'] = 'H:i';
```
capture d ecran 2017-06-29 a 19 08 08
capture d ecran 2017-06-29 a 19 20 41

with 
```
        $graphgenerator->settings['timeScale'] = 'PT1H';
        $graphgenerator->settings['timeGroupFormat'] = 'Y-m-dTH:00:00';
        $graphgenerator->settings['timeFormat'] = 'H:00';
```
capture d ecran 2017-06-29 a 19 07 34

In fact, it will be nice to have the first graph with the time legend of the second.  😎

Skywalker-11:
Problem is if we still use minutes for multiple hours or even seconds this will result in a much greater performance  hit on the script and I don't think that it is realy useful to see a minute based chart there. You won't be able to see the difference between the entries for minutes.

stefaweb:
@Skywalker-11 

If you have a too small sampling (1 hour), the graph is crushed and we lost detail. For me, sampling should stay at 1 minute (or 5 minutes as in monitoring tools). Only the time range to show should change.

Sample of one hour is Ok for week, month or year period but currently, we don't manage this kind of period in the Traffic Graph.

I tested the 1 minute sampling with a 680 000 records database. No load, no cpu stress. Have look at the last screen shot I posted on #848. It show your that one minute sampling is ok but not one hour.

Skywalker-11:
I checked again and your right. I tested it before and it was 1-2 seconds slower with minute based graph but now it works fine without difference in duration.

mailscanner/lib/lineConfig.js
```diff
@@ -94,6 +94,7 @@ function printLineGraph(chartId, settings) {
               showLine: (typeof settings.types === 'undefined' || settings.fillBelowLine[i] ? true :
                           (typeof settings.drawLines === 'undefined' ? false : settings.drawLines)),
               pointRadius: 1,
+              borderWidth: 0.1,
```
stefaweb: Unfortunately, it's not good.

capture d ecran 2017-07-03 a 11 53 38

The problem should come from another place as with PT1H we don't have the problem.

capture d ecran 2017-06-29 a 19 07 34

Skywalker-11: Probably have to adjust the borderwidth a bit more

Skywalker-11: But it actually shows pretty good that you only have some minutes were you are receiving mails so you can say it is at is expected 😁

stefaweb: Tried. No found a good setup with borderWidth.

Seems better with `pointRadius: 0,`

stefaweb: From my prod server with `pointRadius: 0,`.

capture d ecran 2017-07-03 a 12 12 53

PR #887, opened 2017-07-01
Corrected test error in mailwatch_geoip_update.php

stefaweb:


tools/Cron_jobs/mailwatch_geoip_update.php
```diff
@@ -170,7 +170,7 @@
 
         // Apply MailWatch rights on files from the last run
         $mwUID =  exec('cat /etc/sudoers.d/mailwatch | grep "User_Alias MAILSCANNER" | sed "s/.*= \(.*\).*/\1/"', $retval_cat);
-        if ($retval_cat > 0) {
```
endelwar: $retval_cat is an INT, why do you check as an array?
http://php.net/manual/en/function.exec.php

stefaweb: I've an error with `$retval_cat > 0`. The system say that the result is an array.

endelwar: ok, the second parameter is missing ($output) which is an array, $retval_cat must be the third

endelwar:
```php
$mwUID =  exec('cat /etc/sudoers.d/mailwatch | grep "User_Alias MAILSCANNER" | sed "s/.*= \(.*\).*/\1/"', $output_cat, $retval_cat);
```

Issue #779, opened 2017-05-08
Emoji charachters in subject crash database logging


armorharmer:
### Issue summary
It seems that the problem is when someone puts emoji characters into subject. Mailwatch will stop writing to database. 

Logs will only say something like this:
May  6 14:13:58 mailsecurity MailScanner[64217]: Could not use Custom Function code MailScanner::CustomConfig::InitMailWatchLogging, it could not be "eval"ed. Make sure the module is correct with perl -wc (Error: DBD::mysql::st execute failed: Incorrect string value: '\xF0\x9F\x92\x9F O...' for column 'subject' at row 1 at /usr/share/MailScanner/perl/custom/MailWatch.pm line 189, <CLIENT> line 76.

I fixed it by changing:
from: my $subject = fix_latin($message->{utf8subject});
to: my $subject = $message->{subject};

in /usr/share/MailScanner/perl/custom/MailWatch.pm file.

But now I can see only question marks instead of utf8 characters, altough the mails are being logged into db.

### Steps to reproduce
1. Send mail with emojis in subject (something like this: 😀 😃 😄 😁 😆 😅 😂 )

### Expected result
Mail should be written into DB

### Actual result
Mailwatch stops writing to the database

### Version and method
 - MailWatch Version: [ stable 1.2.2 ]
 - Install type: [ Zip/tarball install ]
 - Updated from an older MailWatch or fresh install: [ Fresh instal ]

### Server configuration
 - Operation System: [ CentOS 6.9 ]
 - PHP version: [ 5.6 ]
 - Database: [ MySQL 5.74 ]
 - Mail Server: [ Postfix ]
 - Used web server: [ Apache 2.2 ]

### Client configuration 
 - Operation System: [ Arch Linux ]
 - Browser: [ Firefox ]

armorharmer:
As I said:
"But now I can see only question marks instead of utf8 characters, altough the mails are being logged into db."
You are correct, but that is the only way that I found that doesn't prevent writing to DB.

stefaweb:
Could you post the complete Subject before encoding?
The one, you got in MTA log.

armorharmer:
May  8 10:14:37 mailscanner MailScanner[13892]: MailWatch: Logging message 89DE140912.A7DB2 to SQL
[...]

PR #889, opened 2017-07-03
fix upgrade.php

Skywalker-11:


upgrade.php
```diff
@@ -244,7 +244,7 @@ function getSqlServer()
     //test if mysql or mariadb is used.
     $sql = 'SELECT VERSION() as version';
     $result = $link->query($sql);
-    if (strpos($result->fetch_array['version'], 'MariaDB') === false) {
+    if (strpos($result->fetch_array()['version'], 'MariaDB') === false) {
```
endelwar: This syntax won't work on php 5.3 (we (try to) will keep compatibility with 5.3 for 1.2)

Skywalker-11: Will it work with `$result->fetch_array()[0]['version']` or what is the problem here?

stefaweb: I've read somewhere here in the past that some part of the code already need 5.4.

Skywalker-11: Yes #779 mentions that we probably need 5.4 for unicode support

endelwar: It will exit with error `Parse error: syntax error, unexpected '[', expecting ',' or ';'`

see this test 3v4l: https://3v4l.org/DXgmp

As for php 5.4 minimum version is related to a better utf8 decoding, on php 5.3 UTF8 decoding works, but sometimes gives strange, but not-blocking, results

Skywalker-11: Your test code is invalid: [https://3v4l.org/93J9i](https://3v4l.org/93J9i)

endelwar: both our codes are valid (direct echo vs variable assignment and echo variable), but substance doesn't change: on php 5.3 you can't access the array returned by a method directly, you need to assign the method result to a temp variable and then use that temp variable (it's array dereferencing, introduced in PHP 5.4)

endelwar: This way https://3v4l.org/GkPYX is compatible with php 5.0.0 and above

Skywalker-11: https://3v4l.org/mCOOB this also

endelwar: schermata 2017-07-03 alle 22 44 58
No, it doesn't run on php less than 5.4

from the php manual (http://php.net/manual/en/language.types.array.php#example-62)

> As of PHP 5.4 it is possible to array dereference the result of a function or method call directly. Before it was only possible using a temporary variable.
> 
> As of PHP 5.5 it is possible to array dereference an array literal.

stefaweb: 😎

Skywalker-11: Sorry their mobile page doesn't show that so I assumed that there is no error

PR #897, opened 2017-07-12
fix deprecated split function (php 7.0)

Skywalker-11:


stefaweb:
Have a look in functions.php, f-prot.awk and sophos.awk. More split() in there. 

functions.php:    $validLang = split(',', USER_SELECTABLE_LANG);
f-prot.awk:  split($1, array, " ");
sophos.awk:/Data file name/ { i = split($2, array, "/"); v_filename = array[i] }

Skywalker-11:
HTMLPurifier defines its own split function in line 12417. It does not use the native php split

stefaweb:
Great then!

Issue #929, opened 2017-08-30
Wrong print in detail.php for B/L and W/L


stefaweb:
In last develop.

The space is not good when using other language than English.

![capture d ecran 2017-08-30 a 11 31 06](https://user-images.githubusercontent.com/12907102/29865971-447412a6-8d77-11e7-901d-20a2b2db48a2.png)

![capture d ecran 2017-08-30 a 11 30 53](https://user-images.githubusercontent.com/12907102/29865972-4493b1e2-8d77-11e7-9db0-c88782e298b4.png)

Skywalker-11:
Can you adding in css `white-space: nowrap;`
 [https://www.w3schools.com/cssref/pr_text_white-space.asp](https://www.w3schools.com/cssref/pr_text_white-space.asp)

stefaweb:
This works (at the end of style.css)

```
a {
    white-space: nowrap;
}
```

stefaweb:
Added PR #931.

stefaweb:
Done.

PR #931, opened 2017-08-30
Add nowrap in detail.php (#929)

stefaweb:


mailscanner/style.css
```diff
@@ -999,3 +999,6 @@ table.mail>tbody>tr>th.alignleft {
     text-align: left;
 }
 
+a {
```
endelwar: Can you limit this css rule to that detail.php specific case?

stefaweb: Hello @endelwar!
A way to to this?

Skywalker-11: You can add a class to the `<a href="..." class="whatever"` and apply it only to that class

stefaweb: Done.

PR #944, opened 2017-09-03
Fixes Issue #943

### LINKED ISSUE #943: Absolute paths to logo files do not resolve without dns record
### Steps to reproduce
1.  Install MailWatch 1.2.5 
2.  Do not configure upstream dns for MailWatch

### Expected result
Logos should render regardless of presence of dns record.

### Actual result
Logos fail to load, this is a result of the addition of MAILWATCH_HOSTURL in commit d0d9595440d3c34e32878451eace6f646e02a82d.  This should be a relative path instead of an absolute path, as not all installations may have an upstream dns record (such as test environments or special use cases of MailWatch).

### Version and method
 - MailWatch Version: [ stable 1.2.5 master ]
 - Install type: [ GitHub checkout ]
 - Updated from an older MailWatch or fresh install: [ Fresh install ])

### Server configuration
 - Operation System: [ CentOS ] version [ 6.9 ]
 - PHP version: [ 5.3 ]
 - Database: [  MariaDB ] version [ 10x ]
 - Mail Server: [ Postfix  ]
 - Used web server: [ Apache ] version [ 2.4 ]

shawniverson:
Uniform use of IMAGES_DIR as a relative path instead of an absolute path.  Fixes #943 .

### DIFF
--- a/mailscanner/auto-release.php
+++ b/mailscanner/auto-release.php
@@ -97,7 +97,7 @@ if (file_exists('conf.php')) {
 </head>
 <body class="autorelease">
 <div class="autorelease">
-    <img src="' . MAILWATCH_HOSTURL . IMAGES_DIR . MW_LOGO . '" alt="' . __('mwlogo99') . '">
+    <img src=".' . IMAGES_DIR . MW_LOGO . '" alt="' . __('mwlogo99') . '">
     <div class="border-rounded">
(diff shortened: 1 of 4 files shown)


shawniverson:
See a deeper issue here...improving fix and rebasing....

PR #945, opened 2017-09-04
Fixes #940

### LINKED ISSUE #940: Unsecure user management for domain admins
Any domain admin without user part (before @) in username can:

- create new and modify domain admin and user acounts without user part(before @)
- change properties of admin account ( ie. downgrade type from admin to domain admin)
- delete admin account

Any domain admin EVEN with full username(user@domain.tld) can create new user with username @domain.tld and subsequently escalate privileges to other domains.

### Version and method
 - MailWatch Version: stable 1.2.5
 - Updated from an older MailWatch or fresh install: Fresh install

### Server configuration
- Operation System: gentoo
- Postfix Version: 3.1.6 
- MailScanner Version: 5.0.3
- SpamAssassin Version: 3.4.1 
- PHP Version: 7.1.8
- MySQL Version: 10.1.26-MariaDB

### Client configuration 
any combination

asuweb:
Fixes issue where domain admins can modify Administrator accounts
Fixes issue where domain admins can view and edit non-related admin accounts

### DIFF
--- a/mailscanner/user_manager.php
+++ b/mailscanner/user_manager.php
@@ -689,11 +689,11 @@ if ($_SESSION['user_type'] === 'A' || $_SESSION['user_type'] === 'D') {
     if ($_SESSION['user_type'] === 'D') {
         if ($_SESSION['domain'] === '') {
             //if the domain admin has no domain set we assume he should see only users that has no domain set (no mail as username)
-            $domainAdminUserDomainFilter = 'WHERE username NOT LIKE "%@%"';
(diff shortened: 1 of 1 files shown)


Skywalker-11:
This won't help in case someone changes the post variables

asuweb:
@Skywalker-11 - true, but it's better than nothing.  For a post hack, they'd need to first find the other variables, such as user name etc, before it would have effect.  

It's not a perfect fix, but better than what's currently there.

PR #948, opened 2017-09-06
Adds check for path of postconf, exim and sendmail executables in sf_version.php

### LINKED ISSUE #947: Postfix Version
Postfix Version stays empty in **sf_version.php** because of wrong path with FreeBSD

non-working
```
    // Add test for MTA
    $mta = get_conf_var('mta');
    if (get_conf_var('MTA', true) === 'postfix') {
        echo '<br>' . "\n";
        echo 'Postfix ' . __('version11') . ' ';
        passthru("/usr/sbin/postconf -d | grep 'mail_version =' | cut -d' ' -f3");
        echo '<br>' . "\n";
    }
```
working
```
    // Add test for MTA
    $mta = get_conf_var('mta');
    if (get_conf_var('MTA', true) === 'postfix') {
        echo '<br>' . "\n";
        echo 'Postfix ' . __('version11') . ' ';
        passthru("/usr/local/sbin/postconf -d | grep 'mail_version =' | cut -d' ' -f3");
        echo '<br>' . "\n";
    }
```

asuweb:
Fixes #947

### DIFF
--- a/mailscanner/sf_version.php
+++ b/mailscanner/sf_version.php
@@ -100,13 +100,15 @@ if ($_SESSION['user_type'] !== 'A') {
     if (get_conf_var('MTA', true) === 'exim') {
         echo '<br>' . "\n";
         echo 'Exim ' . __('version11') . ' ';
-        passthru("/usr/sbin/exim -bV | grep 'Exim version' | cut -d' ' -f3");
+        exec("which exim", $exim);
+        passthru("$exim[0] -bV | grep 'Exim version' | cut -d' ' -f3");
         echo '<br>' . "\n";
     }
     if (get_conf_var('MTA', true) === 'sendmail') {
         echo '<br>' . "\n";
         echo 'Sendmail ' . __('version11') . ' ';
-        passthru("/usr/sbin/sendmail -d0.4 -bv root | grep 'Version' | cut -d' ' -f2");
+        exec("which sendmail", $sendmail);
+        passthru("$sendmail[0] -d0.4 -bv root | grep 'Version' | cut -d' ' -f2");
         echo '<br>' . "\n";
     }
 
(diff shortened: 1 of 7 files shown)


mailscanner/sf_version.php
```diff
@@ -93,7 +93,8 @@
     if (get_conf_var('MTA', true) === 'postfix') {
         echo '<br>' . "\n";
         echo 'Postfix ' . __('version11') . ' ';
-        passthru("/usr/sbin/postconf -d | grep 'mail_version =' | cut -d' ' -f3");
+        exec("which postconf", $postconf);
```
stefaweb: Hello @asuweb!

You can may be mod the others tests in sf_version.php with "which" command?

asuweb:
Yes, that's not a bad idea

mailscanner/sf_version.php
asuweb: Yes, that's not a bad idea as it would prevent any other path issues that have yet to be discovered

mailscanner/sf_version.php
```diff
@@ -93,7 +93,8 @@
...
+        passthru("$postconf[0] -d | grep 'mail_version =' | cut -d' ' -f3");
```
endelwar: You shoud add a check in case the exec() above doesn't find the command
```php
if (isset($postconf[0])) {
passthru("$postconf[0] -d | grep 'mail_version =' | cut -d' ' -f3");
} else {
echo 'postconf not found';
}
```

asuweb: Yes, I could do - although in that case it will fail as it did previously and just not print the version - but you are right, it would be better to display an error rather than a blank :)

asuweb: Done.  I was just being lazy ;)

endelwar:
You can also use `notfound06` for translation of `notfound11`.
In a separate PR these general translation (like 'not found', 'error', etc..) could be moved in a single location removing duplication

asuweb:
Yes, I hadn't searched for other not found strings - I'll change it.

asuweb:
And yes, removing duplication is a worthwhile task - I'll look at that this afternoon

PR #961, opened 2017-09-27
add IMAP auth support

akoscomp:
I add code for IMAP auth support via php imap_open function.

### DIFF
--- a/mailscanner/checklogin.php
+++ b/mailscanner/checklogin.php
@@ -57,18 +57,7 @@ if (
     $_SESSION['user_ldap'] = true;
     $myusername = safe_value($myusername);
     $mypassword = safe_value($mypassword);
-} else {
-    $_SESSION['user_ldap'] = false;
-    if ($mypassword !== '') {
-        $myusername = safe_value($myusername);
-        $mypassword = safe_value($mypassword);
-    } else {
-        header('Location: login.php?error=emptypassword');
-        die();
-    }
-}
-
-if (
+} elseif (
     (USE_IMAP === true) &&
     (($result = imap_authenticate($myusername, $mypassword)) !== null)
 ) {
@@ -76,6 +65,7 @@ if (
     $myusername = safe_value($myusername);
     $mypassword = safe_value($mypassword);
 } else {
+    $_SESSION['user_ldap'] = false;
     $_SESSION['user_imap'] = false;
     if ($mypassword !== '') {
         $myusername = safe_value($myusername);
--- a/mailscanner/checklogin.php
+++ b/mailscanner/checklogin.php
@@ -68,6 +68,24 @@ if (
     }
 }
 
+if (
+    (USE_IMAP === true) &&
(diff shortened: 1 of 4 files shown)


mailscanner/functions.php
```diff
@@ -2972,6 +2972,46 @@ function ldap_authenticate($username, $password)
...
+ * @param string $username
+ * @param string $password
+ * @return null|string
+ */
+function imap_authenticate($username, $password)
+{
+    $username = strtolower($username);
+    $email = $username;
+
+    if (!filter_var($email, FILTER_VALIDATE_EMAIL)) {
```
endelwar: You can use `$username` directly without creating `$email` var

mailscanner/functions.php
```diff
@@ -2972,6 +2972,46 @@ function ldap_authenticate($username, $password)
...
+        $mbox = imap_open(IMAP_HOST, $username, $password, NULL, 0);
+
+        if ( false === $mbox ) {
+            //auth faild
+            return null;
+        }
+
+        $sql = sprintf('SELECT username FROM users WHERE username = %s', quote_smart($email));
+        $sth = dbquery($sql);
+        if ($sth->num_rows === 0) {
```
endelwar: Add a `IMAP_AUTOCREATE_VALID_USER` boolean constant to enable this behaviour, set it false by default

mailscanner/checklogin.php
```diff
@@ -68,6 +68,24 @@
     }
 }
 
+if (
```
endelwar: this `if` should be chained with the one on line 53, using an `elseif`, and `else` on line 60 should also set `$_SESSION['user_imap'] = false;`

Skywalker-11:
You should also add a check if user_imap is used in the user_manager.php:
https://github.com/mailwatch/MailWatch/blob/0140fefb15ea90d68803d52ce124c4b9339f3607/mailscanner/user_manager.php#L781

Skywalker-11:
Maybe we can also make the parameters optional so users don't necessarily need to update their config by changing
```
} elseif (
     (USE_IMAP === true) &&
     (($result = imap_authenticate($myusername, $mypassword)) !== null)
 ) {
``` 
to 
```
} elseif (
     defined('USE_IMAP') &&
     (USE_IMAP === true) &&
     (($result = imap_authenticate($myusername, $mypassword)) !== null)
 ) {
```
And moving `USE_IMAP` and `IMAP_HOST` from the $needed array to the optional array eg:
```
...
  'ENABLE_SUPER_DOMAIN_ADMINS' => array('description' => 'allows domain admins to change domain admins from the same domain'),
  'USE_IMAP' => array('description' => 'use IMAP for user authentication'),
  'IMAP_HOST' => array('description' => 'IMAP host to be used for user authentication')
);
```

endelwar:
I've added @Skywalker-11 suggestions to the code, this need to be tested, specially the optional autocreate user from imap login

Issue #970, opened 2017-11-07
mailwatch shows no clamav status


philrandal:
### Issue summary
Mailwatch doesn't report ClamAV status

### Steps to reproduce
1.  as an administrator, visit clamav_status.php
2. 
3. 

### Expected result
ClamAV version and database version should be reported

### Actual result
Blank output

### Version and method
 - MailWatch Version: develop
 - Install type: Installer (beta)
 - Updated from an older MailWatch or fresh install: Upgrade (from version 1.2.6)

### Server configuration
 - Operation System: CentOS version 7.4
 - PHP version:  5.4 
 - Database: MariaDB 
 - Mail Server: Postfix
 - Used web server: Apache

Running 

# /usr/lib/MailScanner/wrapper/clamav-wrapper  -V
/opt/mailwatch/public/auto-release.php: OK
/opt/mailwatch/public/bayes_info.php: OK
/opt/mailwatch/public/checklogin.php: OK
/opt/mailwatch/public/clamav.awk: OK
/opt/mailwatch/public/clamav_status.php: OK
/opt/mailwatch/public/conf.php.example: OK
/opt/mailwatch/public/database.php: OK
/opt/mailwatch/public/detail.php: OK
/opt/mailwatch/public/docs.php: OK
/opt/mailwatch/public/do_message_ops.php: OK
/opt/mailwatch/public/favicon.ico: OK
/opt/mailwatch/public/filter.inc.php: OK
/opt/mailwatch/public/f-prot.awk: OK
/opt/mailwatch/public/f-prot_status.php: OK
/opt/mailwatch/public/f-secure.awk: OK
/opt/mailwatch/public/f-secure_status.php: OK
/opt/mailwatch/public/functions.php: OK
/opt/mailwatch/public/geoip_update.php: OK
/opt/mailwatch/public/graphgenerator.inc.php: OK
/opt/mailwatch/public/index.php: OK
/opt/mailwatch/public/lists.php: OK
/opt/mailwatch/public/login.function.php: OK
/opt/mailwatch/public/login.php: OK
/opt/mailwatch/public/logout.php: OK
/opt/mailwatch/public/mailq.php: OK
/opt/mailwatch/public/mcafee.awk: OK
/opt/mailwatch/public/mcafee_status.php: OK
/opt/mailwatch/public/mcp_rules_update.php: OK
/opt/mailwatch/public/msconfig.php: OK
/opt/mailwatch/public/ms_lint.php: OK
[...]

PR #972, opened 2017-11-08
Workaround for 'broken' MailScanner clamav-wrapper

philrandal:
Return ClamAV version details when MailScanner's clamav-wrapper swallows the -V parameter because it is expecting a path to scan

Fixes Issue #970

### DIFF
--- a/mailscanner/clamav_status.php
+++ b/mailscanner/clamav_status.php
@@ -45,7 +45,7 @@ if ($_SESSION['user_type'] !== 'A') {
     echo '<td align="center">';
 
     // Output the information from the conf file
-    passthru(get_virus_conf('clamav') . ' -V | awk -f ' . __DIR__ . '/clamav.awk');
+    passthru(get_virus_conf('clamav') . ' . -V | awk -f ' . __DIR__ . '/clamav.awk');
 
     echo '</td>';
     echo '</tr>';


Skywalker-11:
@philrandal the problem in the mentioned issue is that MailScanner resets the permissions of the temporary folder after installation/update so that the webserver (apache/nginx) can't create a folder there which is required to run the clamav-wrapper atm. 
So not sure if this patch solves a problem.

The wrapper script also runs without the patch on CentOS7 and Debian9 with correct permissions on that folder.

philrandal:
Sorry, but I am into defensive programming. I had the problem on CentOS 7, my patch fixes it for me. As the code is only getting ClamAV version info, it's a bit of a no-brainer to make it work regardless.

Skywalker-11:
@philrandal Which MailScanner version are you running?

philrandal:
Whatever was the latest release at the time I filled the bug.

Cheers, Phil

philrandal:
I also got the the same output from the script on an old MailScanner box but MailWatch 1.2.0 did the right thing. I think this is actually a regression in MailWatch. Can't check as I am neither at work or at home.

endelwar:
> I think this is actually a regression in MailWatch

The system call hasn't changed from 1.2.0-beta4 (the [second commit](https://github.com/mailwatch/MailWatch/blame/c0e12958122252017e88de5b58fcae65ac38fff8/mailscanner/clamav_status.php) on MailWatch git history). On my pc I have an archived 1.0.3 version (file created on 10/10/2005) that show the same wrapper call.

schermata 2017-11-27 alle 18 49 42

I didn't looked up at how the clamav wrapper in MailScanner changed from august 2013 till now: if now the path is mandatory to get the right result there should be a check on MailScanner version and do the right system call.

@philrandal, on my mailscanner 5 test machine, the file is owned by group `mtagroup` which is the [common group](https://github.com/MailScanner/v5/search?utf8=%E2%9C%93&q=mtagroup) for all daemons working together to make MailScanner functional, is maybe really a permission problem that cause this misbehaviour?

Skywalker-11:
History of the MS wrapper-script:
wrapper only changed the way user/group are used there since v5 release 
https://github.com/MailScanner/v5/commits/master/common/usr/lib/MailScanner/wrapper/clamav-wrapper

philrandal:
It's not an ownership issue as far as I can see.

We're just invoking the MailScanner clamav wrapper script to get the version info, not scan anywhere.

My fix fixes it for me, causes no issues with earlier versions of MailScanner (tested it on 4.84 instances as well as 5.0.6).

PR #980, opened 2017-11-21
Fixed undefined offset when mail has no size.

### LINKED ISSUE #979: Undefined offset - functions.php line 1504
I run a customised mailwatch instance, with amongst other things logs SMTP rejects and greylisted mail in the DB for display with Mailwatch.  The actual logging is currently taken care of by a separate deamon.

This causes an undefined offset on status.php due to there being no valid "size" for the mail.

asuweb:
Fixes #979

### DIFF
--- a/mailscanner/functions.php
+++ b/mailscanner/functions.php
@@ -1491,10 +1491,10 @@ function get_disks()
  */
 function formatSize($size, $precision = 2)
 {
-    if ($size === null || $size === '0') {
+    if ($size === null) {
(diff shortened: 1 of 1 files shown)


endelwar:
This bug was introduced by strict comparison (which is good btw), I think that the check for litteral 0 `'0'` should be moved with the numerical 0 check on line 1497

asuweb:
I agree on the strict comparison.  And, yes, I can move it down a line.

PR #990, opened 2017-11-27
more restricted access to library and public available files. 

Skywalker-11:
user should allow overwrite of apache config through .htaccess or apply the permissions in vhost to enable it

* moved js scripts from sub folder "lib" to sub folder "js"
* created .htaccess files to limit access


endelwar:
This is a good move, just add an entry on FAQ page warning about apache 2.2/2.4 configuration and incompatibility between .htaccess directives

Skywalker-11:
done https://github.com/mailwatch/mailwatch-docs/pull/30

Issue #255, opened 2016-06-28
Whitelist/Blacklist - email blacklist is by passed when it is sent with more than 2 normal recipient email!


somono168:
First let me describe about my email environment:
- Supposed that i have domain "me.com" for global email and "int.me.com" for local email.
- and I have a fews mailing list: global@me.com, managers@me.com, branches@me.com.
- I create a blacklist from="default" to="global@me.com", and whitelist from="me.com" to="global@me.com". So the allemail from "int.me.com" can not send email to global@me.com.

The blacklist blocked perfectly with sender test@int.me.com when there are two recipients "global@me.com, test01@me.com". However, the blacklist does not work at all when there are more than two recipients "global@me.com, test01@me.com, test02@me.com".

It look like if there are more than two good recipients among the blocked recipient, the blacklist will not work.

My Question is:
Is there anyway to absolutely block the email even there are some normal recipients are included?

Thank you,
##

asuweb:
Are we sure this is actually a Mailwatch bug and not a mailscanner bug/feature?

I think this is related to the whitelist/blacklist logic in mailscanner.

In Message.pm
 ```
# Do the whitelist check before the blacklist check.
  # If anyone whitelists it, then everyone gets the message.
  # If no-one has whitelisted it, then consider the blacklist.
  $iswhitelisted = 0;
  my $maxrecips = MailScanner::Config::Value('whitelistmaxrecips');
  $maxrecips = 999999 unless $maxrecips;


```

somono168:
Thank asuweb,

I will test again based on your information to know if it really Mailscanner bug/features or a Mailwatch Bug.

Skywalker-11:
This should to be a bug in MailWatch because there are no applied whitelist entries. So it should look into the blacklist and find the block rule. 

Maybe we are matching every subdomain for a white/blacklist entry so that an entry for bla.com also apply to fuu.bla.com, bla.fuu.bla.com etc.?

somono168:
I have tested again with scenario like this:

1. In Mailwatch B/W List:
- from: example.com , to: test@example.com => WhiteList
- from: default , to: test@example.com => BlackList

2. Test sending email from test@gmail.com
- test 1 with 5 recipients: (test@example.com, test@gmail.com, test1@gmail.com, test@yahoo.com, test2@yahoo.com) => the result: the email from test@gmail.com got blacklist in MailWatch
[...]

PR #992, opened 2017-11-27
Fix for #255 (black/whitelist only checking first recipient)

Skywalker-11:
For mails with many recipients this can result in much higher load. This results in 6+ more lookups for each recipient.
I can't tell how good this scales and which number of recipients users are accepting in mails so someone.
Testing with a bulk of 1000 recipients in a mail didn't show any delay though.

### DIFF
--- a/MailScanner_perl_scripts/SQLBlackWhiteList.pm
+++ b/MailScanner_perl_scripts/SQLBlackWhiteList.pm
@@ -251,7 +251,6 @@ sub LookupList {
         return 1 if $BlackWhite->{$i}{$ip1};
         return 1 if $BlackWhite->{$i}{$ip1c};
         return 1 if $BlackWhite->{$i}{'default'};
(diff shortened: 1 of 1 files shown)


MailScanner_perl_scripts/SQLBlackWhiteList.pm
```diff
@@ -246,12 +250,12 @@ sub LookupList {
         return 1 if $BlackWhite->{$i}{$ip2c};
         return 1 if $BlackWhite->{$i}{$ip1};
         return 1 if $BlackWhite->{$i}{$ip1c};
+        return 1 if $BlackWhite->{$i}{'default'};
+        return 1 if $BlackWhite->{$i}{'default'};
```
stefaweb: Lines 253 and 254 seems the same?

Skywalker-11: copy-paste error😄

PR #1014, opened 2017-12-29
Fixes test in src/bootstrap.php for composer install

asuweb:
Fixes test for required vendor files (composer install) - this was throwing a 500 error due to the `require` coming before a test for the file.

Adds script for automated setup of composer


Skywalker-11:
When does your error occur? I just had to install composer via apt and run `composer install` and it showed no such errors (debian 9)

asuweb:
It occurs when you don't install composer or run `composer install`.  

This would only occur when users haven't correctly setup MailWatch, but as we both know, it will happen. The modification simply ensures that the error message is displayed, rather than a 500 server error.

I deliberately didn't setup composer initially, just to see what would happen. 

And the included shell script simply gets composer.phar and runs `composer install` and is *nix platform independant

PR #1018, opened 2018-01-02
Symfony framework

Skywalker-11:
Use the symfony framework for routing and mvc:
 - login is now done by symfony (src/MailWatch/Controller/LoginController.php). Currently only sql, but ldap module exists
 - old pages are called by src/MailWatch/Controller/XXXCompatController.php: `$this->render($compatSrc . $path);` where $path is the location of the php page eg `status.php` or `user_manager.php` 
 - old sites with pager are currently not working
 - static files (css, js, images) are moved to public folder
 - converted pages can be created like src/MailWatch/Controller/LuckController.php
    - urls for controller function is defined by `@Route(...)`
    - html output can be generated in Controllers by 
        * `return $this->render(<path_in_src_resource_folder>, arguments)`; (see LoginController.php)
        * `return new Response("html here");` (see LuckyController.php)


endelwar:
This is a big step forward for the code base!
Are we all ok with Symfony as a framework? If it's ok we can proceed with this PR, fixing what's not working (e.g.: pagination), and develop in a symfony way auth methods; also it has to be checked the token part for auth, I think Symfony already has something working strong and well that can substitute that part.

ghost:
@endelwar You have my vote.  I like what I am seeing here.

asuweb:
I'm happy with it.  

The way I see it, whatever framework we adopt will take some getting used to, as some of us won't have used a particular framework.  But we have to start somewhere, and it looks good to me.

endelwar:
I've just took half an hour to play with this branch.

Dependencies in composer.json are limiting php 7.0 or better, despite the fact that symfony 3.4 is installed. This is very welcome for me, but we should decide if php 5.6 (which is EOL and will receive security support until end of 2018) has to be supported or not.

Linux distro released before 2016 that don't ship php 7 by default need an external repository to get php 5.6 anyway (see centos 7 which ships 5.4).

If external repository is ok, we should also consider symfony 4 which wants php 7.1

asuweb:
Personally, I think we should go with PHP7.1, and Symfony 4.

It's not that much of a hardship to upgrade to PHP 7.x and since MailWatch should only be being installed by sysadmins with root privileges, this shouldn't pose an issue.  As you say with CentOS 7 you'd have to upgrade to php 5.6 anyway, so it makes sense to use php 7

endelwar:
@shawniverson any thought with an eye on eFa?

ghost:
@endelwar I have frozen development of eFa version 3 that uses CentOS 6.9.

I agree with @asuweb that we should just go ahead and move forward to PHP 7.x.  I am currently testing on CentOS 7 using PHP 7.2 from the remi repository for eFa version 4 with good success (and a noticeable performance improvement :) )

stefaweb:
Hello!

If you switch to php7, could you stay with the version released in Debian 9 ? 
It's more conservative but seem more secure than to go to esoteric versions.

ghost:
@stefaweb We need to think about this then, because Symfony 4 requires PHP 7.1.   Debian 9 uses PHP 7.0.

endelwar:
@stefaweb on debian you can use the updated packages made by Ondřej Surý who is in charge of packaging php for debian official releases: you find them on [deb.sury.org](https://deb.sury.org) (this is the first thing that I install on new debian web servers)

Skywalker-11:
We should also keep in mind that the Symphony 4.0 version won't get LTS support and the support ends july 2018 while 3.5 will be supported and maintained till 2020.

On the other hand the code I added to apply the framework was largely from the 4.0 version tutorials/examples which worked in all cases. I actually assumed I was running the 4.0 version till @asuweb's comment and I looked it up again 😁

endelwar:
> We should also keep in mind that the Symphony 4.0 version won't get LTS support and the support ends july 2018 while 3.5 will be supported and maintained till 2020.

Indeed, let's remain on Symfony 3.4 LTS, minimum php version 7.0
We will reconsider upgrading to 4 LTS when/if it will be released somewhere in 2019/2020

stefaweb:
Thanks fort all these feedback to my question.
Have a nice day.

stefaweb:
Hello!

I don't know where to put this.

Beware, I discovered that Doctrine is not compatible with MariaDB 10.2.x. I found this problem with one of our development using symfony.

https://github.com/doctrine/dbal/pull/2825

Cheers,

ghost:
@stefaweb I am running MariaDB 10.1.31 and so far have not seen issues.  Do we need to stick on the 10.1.x to avoid this problem for now?

stefaweb:
Yes, do not use Mariadb 10.2.x + PHP 5.6 with doctrine < 2.7.

We got this:

[Mapping]  OK - The mapping files are correct.
[Database] FAIL - The database schema is not in sync with the current mapping file.

endelwar:
Symfony 3.4 LTS uses doctrine/orm 2.5 and doctrine/dbal 2.5, so no support for MariaDB 10.2 for it.
MariaDB 10.2 support will land in doctrine/orm 2.7 (or 3.0) which is still unreleased and thus not included in symfony 3.4: IMO we have to wait symfony 4 LTS for it

endelwar:
I've manually rebased and merged this PR into develop branch, so now MailWatch is officially a Symfony 3.4 application (well, sort of...) 🎉.

A lot of work has to be done to make the actual code look like a symfony app, but it's a start.

Thanks @Skywalker-11 for your work 🥇!

Issue #1042, opened 2018-02-21
' (apostrophe) in MailWatch username


SelcomSys:
We are using Mailwatch and have a user with an email address that contains an apostrophe ('). I am unable to create the account and get the error: For all users other than Administrator you must use an e-mail address for the username. 

Is there something I am missing?

Skywalker-11:
Can you provide your system setup as described here: https://github.com/mailwatch/MailWatch/blob/1.2/.github/ISSUE_TEMPLATE.md

endelwar:
I think that the actual code that validate and clean the username is too much aggressive, and need to be recalibrated to conform to RFC 5322 and 5321. Valid email address can be 

> #!$%&'*+-/=?^_`{}|~@example.org

 or 

> "()<>[]:,;@\\\"!#$%&'-/=?^_`{}| ~.a"@example.org

On 1.2 we can use a validator like PHP filter_var with FILTER_VALIDATE_EMAIL and for 2.0 a more complex validator like [Zend\Validator](https://docs.zendframework.com/zend-validator/validators/email-address/)

SelcomSys:
Hi Skywalker, I have just heard back from the guy who built the server and he used the VMWare build from the EFA project: https://efa-project.org/download/. I am not sure what version of MailWatch this has bundled in.

Endelwar, I agree that the code is too aggressive. We are going to look at the code here. We are not sure how easy it will be to access considering we have installed a prebuilt image.

shawniverson:
@SelcomSys In your case the fix is simple.  See PR #1043 and edit functions.php accordingly on the appliance.

ghost:
@SelcomSys PR #1043 is WIP at the moment.  Stand by.

SelcomSys:
Thanks for all your help with this everyone! Excellent response.

ghost:
@SelcomSys The code changes appear to be in the 1.2 branch now.

PR #1043, opened 2018-02-23
Email validation for username

ghost:
Fix Issue #1042 

Keep old validation as a secondary for usernames that are not necessarily email addresses.

### DIFF
--- a/mailscanner/filter.inc.php
+++ b/mailscanner/filter.inc.php
@@ -334,14 +334,14 @@ WHERE
         dbconn();
         if (count($this->item) > 0) {
             // Delete the existing first
-            $dsql = "DELETE FROM `saved_filters` WHERE `username`='" . $_SESSION['myusername'] . "' AND `name`='$name'";
+            $dsql = "DELETE FROM `saved_filters` WHERE `username`='" . safe_value(stripslashes($_SESSION['myusername'])) . "' AND `name`='$name'";
             dbquery($dsql);
             foreach ($this->item as $key => $val) {
                 $sql = "REPLACE INTO `saved_filters` (`name`, `col`, `operator`, `value`, `username`)  VALUES ('$name',";
                 foreach ($val as $value) {
-                    $sql .= "'$value',";
+                    $sql .= "'" . safe_value($value) . "',";
                 }
-                $sql .= "'" . $_SESSION['myusername'] . "')";
+                $sql .= "'" . safe_value(stripslashes($_SESSION['myusername'])) . "')";
                 dbquery($sql);
             }
         }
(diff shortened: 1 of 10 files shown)


Skywalker-11:
Have you tested if there could be problems with the sql queries?
The change also breaks global admins without mail account.

ghost:
"The change also breaks global admins without mail account."

@Skywalker-11 I Where are you seeing this?  I have the code in place on my system and have a global admin that is not a mail account.  The code validates using both methods.

ghost:
Hold on merging this, though.  I did find an issue with the sanitization that also needs addressed.  This alone is not enough to resolve the specific apostrophie issue.  I'll continue with it as soon as I arrive at work.  I'll take a closer look at the sql queries as well that interact with the username.

Skywalker-11:
@shawniverson sorry I overlooked the else statement. So no explicit problem with global admins.

We should also check the perl code (Black list + spam settings) as it could also circumvent the filters depending on how it is escaped in mysql and perl

ghost:
@Skywalker-11 Good idea.

ghost:
Tests done:

- Password reset
- Alternate recipient email addresses honor "'" in address in user management and email release
- Report searching using username/email containing "'"
- MailScanner proper interpretation of escaped "'" in black and whitelists via SQLBlackWhiteList.pm
- MailScanner proper interpretation of escaped "'" of user in SQLSpamSettings.pm
- Successful login with user containing "'"
- Storage of username as an safe (escaped) value containing "'"
- Display of unescaped username to screen
- Edit/Delete/Logout user with "'"
- Change password as user with "'"
- Removal of escapes in audit log report display
- details.php add to whitelist/blacklist email with "'"
- Filters storing and using "'" correctly (bug found and squashed)
- Saved filters properly save and load "'" in value field
- Duplicate user check properly handles "'"
- Domain admin functionality still working with presence of "'" in usernames self & targets
- Quarantine reports handle email addresses with "'" in username
- Turn magic_quotes_gpc on and retain functionality

ghost:
You'll notice the odd construct `safe_value(stripslashes($var))` throughout the code.  This is precautionary and has a normalizing effect on the escaping of single quotes.  Under normal conditions, it will remove the slashes and add them back.  However, it will also guard against a situation in which a $var was not properly escaped elsewhere in the code and add the slashes without the risk of duplicating the effects of `safe_value($var)` and stacking the slashes.

I also think I need to refactor this code to handle magic quotes if it is turned on...

ghost:
It appears that safe_value already takes magic quotes into account.  

Code is ready for review.

PR #1060, opened 2018-04-20
Simplify ClamAV version detection

philrandal:
Use clamscan directly rather than going round the houses using MailScanner's (possibly incorrect) configuration.


philrandal:
Something similar should be merged into the development branch too

PR #1061, opened 2018-04-20
Simplify ClamAV version detection

philrandal:
Detect ClamAV versions by invoking clamscan and not indirectly using MailScanner's configuration

### DIFF
--- a/mailscanner/clamav_status.php
+++ b/mailscanner/clamav_status.php
@@ -44,8 +44,11 @@ if ($_SESSION['user_type'] !== 'A') {
     echo '<tr>';
     echo '<td align="center">';
 
-    // Output the information from the conf file
-    passthru(\MailWatch\Antivirus::getAntivirusConf('clamav') . ' . -V | awk -f ' . __DIR__ . '/clamav.awk');
+    // get version info from clamscan
+    exec('which clamscan', $clamscan);
+    if (isset($clamscan[0])) {
+        passthru("$clamscan[0] -V | awk -f " . __DIR__ . '/clamav.awk');
+    }
 
     echo '</td>';
     echo '</tr>';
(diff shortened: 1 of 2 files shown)


endelwar:
This should be ported to the 2.0 logic where `\MailWatch\Antivirus::getAntivirusConf()` should be modified to detect a local clamav installation and use it directly instead of asking MailScanner for path.

Also `\MailWatch\Translation` has to be used to get translated strings.

thctlo:
This is not working on my debian system.

this works: 
clamd -V| cut -d" " -f2 | cut -d"/" -f1,2
0.100.1/24819

and this works:
clamdscan -V
ClamAV 0.100.1

but there is not clamscan on debian systems. 

This is my change 
diff sf_version.php /opt/mailwatch/mailscanner/sf_version.php
< left working, right not. > 
```
136,138c136,138
<         exec("which clamd", $clamd);
<         if (isset($clamd[0])) {
<             passthru("$clamd[0] -V | cut -d/ -f1 | cut -d' ' -f2");
---
>         exec("which clamscan", $clamscan);
>         if (isset($clamscan[0])) {
>             passthru("$clamscan[0] -V | cut -d/ -f1 | cut -d' ' -f2");
```

philrandal:
apt-get install clamscan 

should fix it on Debian.

clamdscan -V will fail if clamd is not running, which may or may not be what you want.

I personally prefer to keep version info apart from status info, so think that using clamscan instead of clamdscan is preferable.

Cheers,

Phil

thctlo:
Hai Phil, thanks, now its clear, i do run clamav-daemon yes. 
i'll take note of it, its not preffered in my case, because i like, te lesser packages installed the better. 
maybe its an option to test for clamscan, and if its not available try clamd -V

Skywalker-11:
While discussing https://github.com/mailwatch/mailwatch-install-script/issues/56 we noticed that there is a difference in the way how clamav and clamd work. 
Maybe it is required to distinct them to get the version clamav/clamd.

Issue #1073, opened 2018-05-19
RPC function for message detail view not displaying full data with v1.2.9 upgrade


gordoncolyn:
I upgraded from v1.2.6 to v1.2.9 and have found that the view message page does not display all the email data.  It only shows Received on, Received by and Received from details.  All the rest of the message detail from Received via does not show.

When I change my httpd.conf file and point back to the 1.26 data all the info displays.

I have 3 mail servers, 1 mysql server and an http server for clients to login and view quarantined email and reports.  On v1.2.6 everything works fine.  v1.2.9 has the bug.

### Version and method
 - MailWatch Version: [ stable 1.2.9 ]
 - Install type: [ Zip install ]
 - Updated from an older MailWatch or fresh install: [ Upgrade ] (from 1.2.6 to 1.2.9)

### Server configuration
 - Operation System: [ CentOS version 6.9 ]
 - PHP version: [ 5.3  ]
 - Database: [ MySQL version 5.7.22 ]
 - Mail Server: [ Postfix ]
 - Used web server: [ Apache version 2.2 ]

### Client configuration 
 - Operation System: [ windows 10 ] 
 - Browser: [ firefox]

Skywalker-11:
Are there any errors in the logs of the webservers?

gordoncolyn:
No Errors, this is all I see in access log;
xxx.xxx.xxx.xxx - - [02/Jun/2018:09:25:16 +0200] "POST /mailscanner/rpcserver.php
HTTP/1.0" 200 145 "-" "XML-RPC for PHP 3.0.0.beta"

Skywalker-11:
Are the details shown correctly on the view message page wg https://x.x.x.x/detail.php (opened when clicking the info sign in the recent messages/qurantine list?

Or does it only happen for the /viewmail.php?

Can you post a screenshot showing the differences?

davidgknox:
Just finished upgrading to 1.2.9 and expierencing the same issue since the upgrade. Only "Received on", "Received by" and "Received from" are displayed, all other header info is missing when viewing it through the "Recent Massages" or the "Quarantine" list.

gordoncolyn:
I get the same result on the detail page;
![image](https://user-images.githubusercontent.com/31371167/40878667-7a22cb3a-6694-11e8-84dd-a7eb58fcd42f.png)

gordoncolyn:
It should be;
[...]

PR #1074, opened 2018-06-04
Notice in README and CHANGELOG that PHP 5.4 is required

Skywalker-11:
#1073


Skywalker-11:
@endelwar can you take a look at this?

PR #1085, opened 2018-07-11
fix mail filter for normal users - fixes #1084

### LINKED ISSUE #1084: Mail only shown for first recipient
EDIT: For reference initially reported on the mailing list by Jim (Mailwatch-users Digest, Vol 136)

When mail is addressed to multiple addresses say a@example.com, b@example.com, c@example.com (each normal users) the mail is only listed in the gui in context of the first recipient (here a@example.com). For other recipients the mail is not listed when they log in. 
Mail is correctly shown for domain admins of any recipient domain

### Steps to reproduce
1. mail to a@example.com, b@example.com, c@example.com (single mail multi recipient)
2. login as b@example.com
3. mail not listed

v1.2.9 on CentOS6 (PHP5) and Debian 9 (PHP7)

Problem is the sql query for filtering of the mails: `.... WHERE (to_address like 'b@example.com%' OR from_address = 'b@example.com')  ` (wildcard only on end => has to be first address to be listed)

Skywalker-11:


endelwar:
Very good solution!

endelwar (approved):

PR #1105, opened 2018-08-26
Fix long queue id validation

ghost:
Minor fix for postfix long queue id validation

### DIFF
--- a/mailscanner/functions.php
+++ b/mailscanner/functions.php
@@ -4467,7 +4467,7 @@ function validateInput($input, $type)
             }
             break;
         case 'msgid':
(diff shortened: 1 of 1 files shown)


Skywalker-11:
Can you explain where `\.[A-F0-9]{5}` comes from?

ghost:
@Skywalker-11 

https://github.com/MailScanner/v5/blob/master/common/usr/share/MailScanner/perl/MailScanner/Postfix.pm#L1878-L1884

I think this was an oversight when the long queue ID support was added a while back.  It is still appending the 5 character key to the ID.  There should be logic here to only do this to short queue IDs.  Nonetheless, the weird looking long queue ID + 5 char key should be permitted since it has persisted in recent versions of MailScanner.

ghost:
Fixing in my working branch. 

https://github.com/shawniverson/v5/commit/77e4117798747c4b6c94048de66e5bf0ce6a4018

ghost:
Another modification to fix

https://github.com/shawniverson/v5/commit/f1782935727b9247111f0f6246f1d81c6176e4a2

PR #1124, opened 2018-12-25
MailScanner milter relay stats support

ghost:
Add support to capture relay statistics from the mail log when the milter is active.

When the milter is active, it does not touch the postfix queues.  This has the effect of MailScanner not having the ability to set the queue ID directly in postfix when requeuing.

Using the message id instead, it is possible to match up the relay info to the message.

### DIFF
--- a/tools/Postfix_relay/mailwatch_milter_relay.php
+++ b/tools/Postfix_relay/mailwatch_milter_relay.php
@@ -1,0 +1,98 @@
+<?php
+
+/*
+ * MailWatch for MailScanner
+ * Copyright (C) 2003-2011  Steve Freegard (steve@freegard.name)
+ * Copyright (C) 2011  Garrod Alwood (garrod.alwood@lorodoes.com)
+ * Copyright (C) 2014-2018  MailWatch Team (https://github.com/mailwatch/1.2.0/graphs/contributors)
+ *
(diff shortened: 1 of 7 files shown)


tools/Postfix_relay/mailwatch_milter_relay.php
```diff
@@ -0,0 +1,72 @@
...
+    if (!$fp = popen($input, 'r')) {
+        die(__('diepipe54'));
+    }
+
+    $lines = 1;
+    while ($line = fgets($fp, 2096)) {
+        if (preg_match('/^.*postfix\/cleanup.*: (\S+): message-id=(\S+)$/', $line, $explode)) {
+            $smtp_id = $explode[1];
+            $message_id = $explode[2];
+            $smtpd_id = database::mysqli_result(dbquery("SELECT id from `maillog` where headers LIKE '%" . $message_id . "%' LIMIT 1;"));
```
ghost: Would this cause problems on a busy server with a large maillog table?

ghost: lol, I did a query on a maillog with >300,000 rows and it took over 1 minute and pegged a CPU core at 100%.  Answered my own question.  Going to refactor this and try again...

endelwar:
👌 Look good for me!

Issue #1117, opened 2018-12-07
rep_viruses.php


tgreggs:
MailWatch 1.2.10 rep_viruses.php does not include the code for Avast. Same with 1.2.11. Any chance that could be included?

endelwar:
Can you provide a working regex or some meaningful output string from avast client?

tgreggs:
Here is the output from a command line scan with the same scan options used by MailScanner wrapper for Avast:

scan -b -f -a message 

/var/spool/MailScanner/quarantine/20181207/A951D21208AF.A7C41/message   Other:Malware-gen [Trj]

Skywalker-11:
Can you try adding the following at https://github.com/mailwatch/MailWatch/blob/1.2/mailscanner/rep_viruses.php#L108 

```
        case('avast'):
            $scanner[$vscanner]['name'] = 'Avast';
            $scanner[$vscanner]['regexp'] = "/(\S+) (\S+)/";
            break;
```
And at https://github.com/mailwatch/MailWatch/blob/1.2/mailscanner/functions.php#L236 add
```
            case 'avast':
                $regex = '/(\S+) (\S+)/';
                break;
```

Then check if it works correct?

tgreggs:
Here is what the virus report now looks like.

[MailWatch Filter Report_ Virus Report.pdf](https://github.com/mailwatch/MailWatch/files/2663819/MailWatch.Filter.Report_.Virus.Report.pdf)

Skywalker-11:
Can you check if the output for all malware that is found by avast match the pattern `<path> <type> [xxx]` especially the square brackets to distinct them from other messages.

tgreggs:
It kind of seems like the way sophos and avast are logged by MailScanner could be part of the issue.  Am running both.

MailScanner 5.1.1 Postfix/Dovecot
MailWatch 1.2.10

MailScanner uses Sophos and Avast.  Notice in the logging the 1st two lines are Sophos results.  

Maillog:
Dec  4 18:54:23 mail MailScanner[6973]: >>> Virus 'Mal/Fareit-Q' found in file /var/pool/MailScanner/incoming/6973/CA20221208AC.A40F1/z1INVOICES.exe
Dec  4 18:54:23 mail MailScanner[6973]: >>> Virus 'Mal/Fareit-Q' found in file /var/pool/MailScanner/incoming/6973/CA20221208AC.A40F1/nINVOICES.zip/INVOICES.exe
[...]

PR #1133, opened 2019-01-14
use named virus regex; added avast regexp; fix nod32-1.99 virus regex

Skywalker-11:


endelwar:
@Skywalker-11 is this code good to be merged?

Skywalker-11:
It should work but there are cases where it might fail. These cases would probably also exist with the current code (see https://github.com/mailwatch/MailWatch/issues/1117#issuecomment-455282793)

Skywalker-11:
Especially if multiple scanners are used or if a scanner marks multiple viruses in a single scan this introduces problems with this PR and also the current code.
This would require a change in MailScanner https://github.com/MailScanner/v5/issues/355

endelwar:
It contains no harm, a fix and an additional AV, for me is a GO

PR #1171, opened 2020-03-23
Quote handling in mailwatch_milter_relay.php

### LINKED ISSUE #1170: Single quote handling in message ID
### Issue summary
Single quotes need handled properly in the message ID field.

### Steps to reproduce
1. Send an email with a single quote in the message ID

### Expected result
Quote should be escaped before inserting into DB.

### Actual result
https://forum.efa-project.org/viewtopic.php?f=5&t=4220&p=15708

ghost:
Fixes #1170

### DIFF
--- a/tools/Postfix_relay/mailwatch_milter_relay.php
+++ b/tools/Postfix_relay/mailwatch_milter_relay.php
@@ -50,7 +50,7 @@ function doit($input)
     while ($line = fgets($fp, 2096)) {
         if (preg_match('/^.*postfix\/cleanup.*: (\S+): message-id=(\S+)$/', $line, $explode)) {
             $smtp_id = $explode[1];
-            $message_id = $explode[2];
+            $message_id = preg_replace("'", "''", $explode[2]);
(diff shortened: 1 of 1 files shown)


tools/Postfix_relay/mailwatch_milter_relay.php
```diff
@@ -50,7 +50,7 @@ function doit($input)
     while ($line = fgets($fp, 2096)) {
         if (preg_match('/^.*postfix\/cleanup.*: (\S+): message-id=(\S+)$/', $line, $explode)) {
             $smtp_id = $explode[1];
-            $message_id = $explode[2];
+            $message_id = preg_replace("'", "''", $explode[2]);
```
endelwar: I think that a `safe_value()` on both $smtp_id and $message_id must be used, because these 2 values are used in SQL query as is, and are prone to sql inject attack

endelwar:
```php
$smtp_id = safe_value($explode[1]);
$message_id = safe_value($explode[2]);
```

ghost: That makes more sense.

PR #1180, opened 2020-05-06
Fix TLD check in whitelist / blacklist

smyers119:
This should fix the lack of TLD support for the blacklist and whitelist of mailwatch.  regex validated using validator.

mail.example.com
gets broken down to 
mail.google.com
google.com
com
instead of just
mail.google.com
google.com


smyers119:
How many sub-domains do we need to realistically check for?  may want to use something other then regex if we are breaking down more then 3 levels which is what my change does,  Before the regex only broke it down 2 levels.

smyers119:
We could break the TLD into a seperately

PR #1193, opened 2020-08-27
Refactor mailwatch_postfix_relay.php and mailwatch_milter_relay.php to use fopen() instead of popen() with tail

ghost:
On some systems, the call to popen() can leave an orphaned and out of control tail process running when the script is prematurely stopped before pclose() is called to terminate the process. Refactored here to use native php approach using fopen() that avoids the use of a separate tail call.


endelwar:
Look good to me! Thank you Shawn.

PR #1194, opened 2020-09-20
Quarantine reports to include multiple recipients

### LINKED ISSUE #1188: Quarantine Report does not display quarantined spam when message sent to multiple recipients is quarantined as spam
Hi,

The Quarantine Report that is emailed to users does not display the quarantined spam if a message is sent to multiple recipients and is quarantined as spam.  So emails are available to view in the web interface, however the email report does not show these emails as they have multiple recipients.

Server is running mailwatch v1.2.11 installed from zip on centos 7 server, php 7.3, mysql 5.7, postfix, apache

ghost:
Resolve Issue #1188

### DIFF
--- a/mailscanner/quarantine_report.inc.php
+++ b/mailscanner/quarantine_report.inc.php
@@ -460,7 +460,12 @@ ORDER BY a.date DESC, a.time DESC';
     private static function return_quarantine_list_array($to_address, $to_domain)
     {
         if ($to_address != '' || $to_domain != '') {
-            $result = dbquery(sprintf(self::get_report_sql(), quote_smart('%' . $to_address . '%'), quote_smart($to_domain)));
+            # Qualify to email addresses only for wildcard search
+            if (preg_match('/(\S+)@(\S+)/', $to_address)) {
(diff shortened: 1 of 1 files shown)


mailscanner/quarantine_report.inc.php
```diff
@@ -460,7 +460,12 @@ private static function return_user_filters($user)
     private static function return_quarantine_list_array($to_address, $to_domain)
     {
         if ($to_address != '' || $to_domain != '') {
-            $result = dbquery(sprintf(self::get_report_sql(), quote_smart($to_address), quote_smart($to_domain)));
+            # Qualify to email addresses only for wildcard search
+            if (preg_match('/(\S+)@(\S+)/', $to_address)) {
+                $result = dbquery(sprintf(self::get_report_sql(), quote_smart('%' . $to_address . '%'), quote_smart($to_domain)));
```
ghost: Problem here is that abc@example.com will overlap with cabc@example.com.  This is going to need more work.

ghost (commented):
Fixed by using RLIKE instead of LIKE and creating a regex.

endelwar (approved):

PR #1211, opened 2021-03-17
Add new config flag to permit IMAP login without full email as username

### LINKED ISSUE #1210: IMAP Login Failures
### Issue summary
MailWatch configured for IMAP authentication does not work with an IMAP server that does not use the full email address as the login.

### Steps to reproduce
1. Configure IMAP server to use username rather than username@maildomain.tld for the login
2. Configure MailWatch to use IMAP authentication with that IMAP server
3. Attempt to login

### Expected result
User should be logged in stripping the @maildomain.tld if necessary for IMAP authentication (automatically or configurable).  

### Actual result
Using only username:  No attempted login - https://github.com/mailwatch/MailWatch/blob/1.2/mailscanner/functions.php#L3208

Using username@maildomain.tld: imap_open(): Couldn't open stream ...

### Version and method
 - MailWatch Version: 1.2.16
 - Updated from an older MailWatch or fresh install: Fresh install

endelwar:
Fix #1210

### DIFF
--- a/mailscanner/functions.php
+++ b/mailscanner/functions.php
@@ -3206,7 +3206,18 @@ function imap_authenticate($username, $password)
 {
     $username = strtolower($username);
 
-    if (!filter_var($username, FILTER_VALIDATE_EMAIL)) {
+    if (
+        (
+            !defined('IMAP_USERNAME_FULL_EMAIL') &&
+            !filter_var($username, FILTER_VALIDATE_EMAIL)
+        )
+        ||
+        (
+            defined('IMAP_USERNAME_FULL_EMAIL') &&
+            IMAP_USERNAME_FULL_EMAIL === true &&
+            !filter_var($username, FILTER_VALIDATE_EMAIL)
+        )
+    ) {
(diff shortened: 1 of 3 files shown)


mailscanner/functions.php
```diff
@@ -4252,7 +4263,8 @@ function checkConfVariables()
         'ENABLE_SUPER_DOMAIN_ADMINS' => array('description' => 'allows domain admins to change domain admins from the same domain'),
         'USE_IMAP' => array('description' => 'use IMAP for user authentication'),
         'IMAP_HOST' => array('description' => 'IMAP host to be used for user authentication'),
-        'IMAP_AUTOCREATE_VALID_USER' => array('description' => 'enable to autorcreate user from valid imap login'),
+        'IMAP_AUTOCREATE_VALID_USER' => array('description' => 'enable to autocreate user from valid imap login'),
+        'IMAP_USERNAME_FULL_EMAIL' => array('description' => 'set to fals eto permit imap login without valid email as username'),
```
Skywalker-11: typo `fals eto`

PR #1221, opened 2021-08-26
Fix rule_name formatting in sa_rules_update.php

amartin-git:
htmlentities() should only be applied to rule_desc, not rule_name as it breaks the rule descriptions in details.php.


endelwar:
Good catch! Thanks

PR #1260, opened 2022-10-22
improvements for service status display

dneuhaeuser:
When Mailscanner is in Milter mode (MTA = msmail) the service status reports a different process count for postfix.
For consistency I suggest to use the same ps command as without Milter mode.

On linux systems (e.g. Ubuntu 22.04.1 LTS) there is a header line from ps output. 
This should be cut, otherwise postfix is reported as running when it isn't and the process count is too high by 1.

Also, with Mailscanner's "Milter Dispatcher = postfork" and the system idling, the status displays:
"MSMilter: YES | 0 children", which is confusing.
I suggest to display the process count when it is 1.

### DIFF
--- a/mailscanner/functions.php
+++ b/mailscanner/functions.php
@@ -461,7 +461,7 @@ function printServiceStatus()
         // is MTA running
         $mta = get_conf_var('mta');
         if ('msmail' === $mta) {
-            exec('ps ax | grep postfix | grep -v grep | grep -v php', $output);
+            exec('ps -U postfix -u postfix | grep -v MailScanner | grep -v "MailWatch SQL" | grep -v MSMilter | grep -v PID', $output);
             if (count($output) > 0) {
                 $running = $yes;
             } else {
--- a/mailscanner/functions.php
+++ b/mailscanner/functions.php
@@ -461,7 +461,7 @@ function printServiceStatus()
         // is MTA running
         $mta = get_conf_var('mta');
         if ('msmail' === $mta) {
-            exec('ps -U postfix -u postfix | grep -v MailScanner | grep -v "MailWatch SQL" | grep -v MSMilter | grep -v PID', $output);
+            exec('ps ax | grep postfix | grep -v grep | grep -v php', $output);
             if (count($output) > 0) {
                 $running = $yes;
             } else {
@@ -475,12 +475,9 @@ function printServiceStatus()
             exec('ps ax | grep MSMilter | grep -v grep', $output);
             if (count($output) > 0) {
                 $running = $yes;
-            } else {
-                $running = $no;
-            }
-            if (count($output) > 1) {
                 $procs = count($output) - 1 . ' ' . __('children03');
             } else {
+                $running = $no;
                 $procs = count($output) . ' ' . __('procs03');
             }
             echo '    <tr><td>' . 'MSMilter' . __('colon99') . '</td>'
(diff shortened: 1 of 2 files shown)


ghost:
@dneuhaeuser Testing PR, will post results.

mailscanner/functions.php
```diff
@@ -461,7 +461,7 @@ function printServiceStatus()
         // is MTA running
         $mta = get_conf_var('mta');
         if ('msmail' === $mta) {
-            exec('ps ax | grep postfix | grep -v grep | grep -v php', $output);
+            exec('ps -U postfix -u postfix | grep -v MailScanner | grep -v "MailWatch SQL" | grep -v MSMilter | grep -v PID', $output);
```
ghost: If we want all procs, we need to assume a master process is present, maybe add 1 to the proc count if running?

ghost: Maybe something like this?
```
 if (count($output) > 0) {
     $running = $yes;
     $procs = count($output) + 1 . ' ' . __('procs03');
 } else {
     $running = $no;
     $procs =  '0 ' . __('procs03');
 }
```

mailscanner/functions.php
```diff
@@ -475,17 +475,20 @@ function printServiceStatus()
             exec('ps ax | grep MSMilter | grep -v grep', $output);
             if (count($output) > 0) {
                 $running = $yes;
-                $procs = count($output) - 1 . ' ' . __('children03');
             } else {
                 $running = $no;
+            }
+            if (count($output) > 1) {
+                $procs = count($output) - 1 . ' ' . __('children03');
```
ghost: To align with the postfix count, maybe identify all procs, not just children, and simply drop the - 1 on line 478 and change it to `__('procs03')` to keep it simple?

mailscanner/functions.php
ghost: Another option would be to explicitly check for postfix/master proc and add it (I would like this better)

dneuhaeuser: How about this?

```
ps --pid `ps -C master -o pid=` --ppid `ps -C master -o pid=` -o pid=,cmd=
```

This would search for the postfix/master proc, lookup its childs and output both.

sample output:
```
1601555 /usr/lib/postfix/sbin/master -w
1601557 qmgr -l -t fifo -u
1601721 tlsmgr -l -t unix -u -c
2052128 pickup -l -t fifo -u -c -o content_filter= -o receive_override_options=no_header_body_checks
2066535 showq -t unix -u -c
```

I'm not sure if all used ps parameters are also available on BSD. Can someone confirm?

dneuhaeuser: How about something like this?

```
            exec('ps -C master -o pid=',$masterpid);
            if (count($masterpid) > 0) {
                $running = $yes;
                exec('ps --pid '.$masterpid[0].' --ppid '.$masterpid[0].' -o pid=',$childpids);
                $procs = count($childpids) . ' ' . __('procs03');
            } else {
                $running = $no;
                $procs = '0 ' . __('procs03');
            }
```

This would search for the postfix/master proc, lookup its childs and output the total count of both.

I'm not sure if all used ps parameters are also available on BSD. Can someone confirm?

mailscanner/functions.php
dneuhaeuser: I agree, that is more logical.

So just like this?

```
            exec('ps ax | grep MSMilter | grep -v grep', $output);
            if (count($output) > 0) {
                $running = $yes;
            } else {
                $running = $no;
            }
            $procs = count($output) . ' ' . __('procs03');
```

dneuhaeuser:
sorry, did not wanted to put this additional commit here but in a separate PR...

PR #1261, opened 2022-10-26
msre: fix status message after mailscanner reload

dneuhaeuser:
the else case is the good case, so it should give us a positive status message (not an error!) 
'error55' isn't even defined in the language files.


endelwar:
Thanks!

PR #1262, opened 2022-10-26
corrected and improved german translation

dneuhaeuser:
corrected and improved german strings for quarantine_report


endelwar:
Thanks!

PR #1263, opened 2022-10-27
improvements for service status display

dneuhaeuser:
Unfortunately I messed up my first PR, sorry for that!

So here are my further reworked improvements for the service status again.
This should now be compatible with BSD and Linux.

When Mailscanner is in milter mode (MTA = msmail) the service status only counted the postfix master process (so the count was always just showing 1). 
For consistency I suggest to use the same logic as without milter mode and display the total postfix process count including its child procs.

On Linux systems (e.g. Ubuntu 22.04.1 LTS) by default there is a header line in the 'ps' output.
Therefore postfix was reported as running when it wasn't because the process count was always too high by 1.

Also, with Mailscanner's "Milter Dispatcher = postfork" and the system idling, the status displays:
"MSMilter: YES | 0 children", which is confusing.
I suggest to always just display the total process count (and to not subtract 1 for a display of child count).


ghost:
I'll take a look at this and let you know my feedback

mailscanner/functions.php
```diff
@@ -457,37 +457,39 @@ function printServiceStatus()
...
-        } else {
-            $psExecCommand = "ps ax | grep $mta | grep -v grep | grep -v php";
-            if ('postfix' === $mta) {
-                $psExecCommand = 'ps -U postfix -u postfix | grep -v MailScanner | grep -v "MailWatch SQL"';
-            }
-            exec($psExecCommand, $output);
+        }
+        if (('msmail' !== $mta) && ('postfix' !== $mta)) {
+            $output = [];
+            exec('ps ax | grep $mta | grep -v grep | grep -v php', $output);
```
endelwar: It's better to use `sprintf()` to inject variable in a string:
```diff
-exec('ps ax | grep $mta | grep -v grep | grep -v php', $output);
+exec(sprintf('ps ax | grep %s | grep -v grep | grep -v php', $mta), $output);
```

Single quotes here won't work because there will not be variable substitution.

mailscanner/functions.php
```diff
@@ -457,37 +457,39 @@ function printServiceStatus()
...
 
-        $output = 0;
         // is MTA running
         $mta = get_conf_var('mta');
-        if ('msmail' === $mta) {
-            exec('ps ax | grep postfix | grep -v grep | grep -v php', $output);
-            if (count($output) > 0) {
+        if (('msmail' === $mta) || ('postfix' === $mta)) {
+            $masterproc = [];
+            exec('ps ax | grep postfix.*master | grep -v grep',$masterproc);
```
endelwar: Please run `vendor/bin/php-cs-fixer fix` to fix code style

mailscanner/functions.php
dneuhaeuser: right, I missed that one.

mailscanner/functions.php
dneuhaeuser: OK, will do that.

btw: why is php-cs-fixer always failing here on github PRs?

endelwar: > btw: why is php-cs-fixer always failing here on github PRs?

There was a problem with GitHub workflow, it should be fixed on `1.2` branch, try rebasing your PR on that branch.

ghost:
This looks good, I tested it on my instance.

PR #1266, opened 2022-10-31
improve performance of mailwatch_milter_relay.php

dneuhaeuser:
the code part which executes the SQL commands does not need to run after each line from the logfile is processed.
instead it is sufficient to run SQL statements once after logfile is analyzed and relevant data was read into $idqueue array.
so I moved this part into a separate function. 
this is a quite simple but very performance-effective change.

### DIFF
--- a/tools/Postfix_relay/mailwatch_milter_relay.php
+++ b/tools/Postfix_relay/mailwatch_milter_relay.php
@@ -47,7 +47,6 @@ $idqueue = [];
 
 function doit($input)
 {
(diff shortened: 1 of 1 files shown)


ghost:
Very nice! I'll give this a test drive as well

ghost:
This is good.

dneuhaeuser:
can we merge this one or do you want any further changes?

endelwar:
Thanks again!

PR #1268, opened 2022-11-07
enhancement for optionally hiding nonspams on quarantine display

dneuhaeuser:
enhance quarantine.php display to regard all **HIDE_...** configuration options like quarantine_report.inc.php does.

this is especially useful when you have Mailscanner configured like this:
`Non Spam Actions = store-nonspam ...`

Then you can get only the relevant quarantine to display with mailwatch config:
`define('HIDE_UNKNOWN', true);`

### DIFF
--- a/mailscanner/conf.php.example
+++ b/mailscanner/conf.php.example
@@ -298,6 +298,9 @@ define('HIDE_NON_SPAM', false);
 // Hide Unknown Mail from quarantine reports
 define('HIDE_UNKNOWN', false);
 
+// Apply hide options above also to quarantine page
+define('HIDE_APPLY_QUARANTINE', false);
+
 // Quarantine Auto Release
 // Set true to allow auto release of quarantined items from quarantine report.
 define('AUTO_RELEASE', false);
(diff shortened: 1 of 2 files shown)


ghost:
I don't think this is desired behavior for some folks (myself included).  Although I don't want certain things in the quarantine report, I want to be able to see all mail in the UI.  Perhaps need an option to toggle this functionality so that both behaviors can be achieved.

dneuhaeuser:
So, you have **HIDE_UNKNOWN = true** but want to see the **nonspam** mails on the quarantine page?

My users actually found this very confusing. They only go to the quarantine page to release spam or bad mails and expect to see only the mails which where reported by quarantine_report there.

All mails are always visible in message listing and message operations, to inspect non-spams for example.

Where is the point to show non-spams in the quarantine?

ghost:
I agree that may be desirable for users.  I think there should be a setting that enables this desired behavior as an option so that those that don't want that behavior can disable it.

dneuhaeuser:
I added a configuration parameter to enable or disable whether the hide filters are applied to the quarantine page.

endelwar:
Thanks Dennis!

PR #1270, opened 2022-11-08
improve quarantine_report

dneuhaeuser:
Having administrator users defined with one or more filter addresses (e.g. automatically imported from LDAP with mailwatch_ldap_sync.sh) led to quarantine reports with duplicate lines.

So when the user type is 'A' we must NOT iterate through the filters because administrators get a report for every single recipient anyway.

Also the administrator's report had an incomplete headline because a suitable string was missing ($list_for was set empty). 
I suggest using the system's hostname in this case.

For normal users, when sending **combined** reports the headline should only list the addresses which are contained in the current report. Otherwise this gets very messy when you have a lot of address filters defined for a user.


ghost:
Another great PR, thanks!  I'll have some time this coming weekend to go over things.

ghost:
This one looks good, tested good as well.

PR #1281, opened 2023-07-08
enhancement for individual nonspam quarantine keeping

dneuhaeuser:
I found it useful to have quarantine keeping days independently configured for **nonspam** mails.

Of course this is only relevant if you also store nonspams like this:
`Non Spam Actions = store-nonspam ...`

So with this enhancement you can for example keep the spams for 30 days but the nonspams only for let's say 10 days. 
On busy systems this can save performance and drive space.

The other way around is also possible if needed: e.g. keep spam 20 days and nonspam 40 days.

### DIFF
--- a/tools/Cron_jobs/mailwatch_quarantine_maint.php
+++ b/tools/Cron_jobs/mailwatch_quarantine_maint.php
@@ -66,7 +66,8 @@ if (0 === $required_constant_missing_count) {
 
     function quarantine_clean()
     {
-        $oldest = date('U', strtotime('-' . QUARANTINE_DAYS_TO_KEEP . ' days'));
+        $oldestspam = date('U', strtotime('-' . QUARANTINE_DAYS_TO_KEEP . ' days'));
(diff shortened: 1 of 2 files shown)


tools/Cron_jobs/mailwatch_quarantine_maint.php
```diff
@@ -66,16 +66,26 @@ function quarantine_reconcile()
 
     function quarantine_clean()
     {
-        $oldest = date('U', strtotime('-' . QUARANTINE_DAYS_TO_KEEP . ' days'));
+        $oldestspam = date('U', strtotime('-' . QUARANTINE_DAYS_TO_KEEP . ' days'));
+        $oldestnonspam = date('U', strtotime('-' . QUARANTINE_DAYS_TO_KEEP_NONSPAM . ' days'));
```
endelwar: There should be a check here to confirm if constant exists, or fallback on `QUARANTINE_DAYS_TO_KEEP`

dneuhaeuser: sure, I will add that...

PR #1283, opened 2023-07-23
Update functions.php to hide snapd mount point

slsguerra:
That's what the title says

image


endelwar:
Thanks @HierraStrunger !

PR #1289, opened 2023-10-14
Update geoip_update.php

slsguerra:
Permalinks from maxmind website says:
/app/geoip_download?edition_id=GeoLite2-Country&license_key=YOUR_LICENSE_KEY&suffix=tar.gz


ghost:
@HierraStrunger Are you seeing an issue with geoip_update.php?  The order of the URL parameters should not matter.

slsguerra:
Yes, the message error is:

Error occurred while downloading GeoIP data file: Response could not be parsed

And the file is not downloaded. After change geoip_update.php, it works

endelwar:
I've fixed in Request library the regex that parses HTTP/2 header, which presents a trailing space not followed by text like in HTTP 1.x

PR #1302, opened 2024-07-04
Update GeoIP to support authentication by account id+license key

### LINKED ISSUE #1301: Cannot Update GeoIP Database
### Issue summary
Cannot update GeoIP Database

On page geoip_update.php, show success download, but no database downloaded when I check on Software Version page

Geoip_update.php :
Downloading file, please wait...
Download complete, unpacking files...

Software Version Page :
GeoIP Database Version: No database downloaded

I check on directory, show like this, size file just 23 bytes, there are no mmdb geolite files

root@rz1:/opt/mailwatch# ls -ltrh /opt/mailwatch/mailscanner/temp/
total 4.0K
-rw-r--r-- 1 www-data www-data 23 Jul  4 22:20 GeoLite2-Country.tar.gz

I've tried changing the maxmind key, but the geolite file is not downloaded

### Version and method
 - MailWatch Version: 1.2.23
 - Updated from an older MailWatch or fresh install: Upgrade from version 1.2.21

### Server configuration
 - Operation System: Ubuntu 22.04.4 LTS (Jammy Jellyfish)
 - PHP version: 8.1.29
 - Database: MariaDB 10.11.8-MariaDB-ubu2204-log
 - Mail Server: Postfix
 - Used web server: Nginx nginx/1.26.1

### Client configuration 
 - Operation System: MacOS 14.5
 - Browser: Chrome 126.0.6478.127

Skywalker-11:
fixes #1301

### DIFF
--- a/mailscanner/geoip_update.php
+++ b/mailscanner/geoip_update.php
@@ -1,6 +1,6 @@
 <?php
 
-/*
+/**
  * MailWatch for MailScanner
  * Copyright (C) 2003-2011  Steve Freegard (steve@freegard.name)
  * Copyright (C) 2011  Garrod Alwood (garrod.alwood@lorodoes.com)
@@ -33,7 +33,7 @@ require __DIR__ . '/login.function.php';
 
 html_start(__('geoipupdate15'), 0, false, false);
 
-if (!defined('MAXMIND_LICENSE_KEY') || !validateInput(MAXMIND_LICENSE_KEY, 'maxmind')) {
+if (!defined('MAXMIND_LICENSE_KEY') || !defined('MAXMIND_ACCOUNT_ID') || !validateInput(MAXMIND_LICENSE_KEY, 'maxmind')) {
     $error_message = __('geoipnokey15') . '<br>' . "\n";
     exit($error_message);
 } elseif (!isset($_POST['run'])) {
@@ -55,15 +55,13 @@ if (!defined('MAXMIND_LICENSE_KEY') || !validateInput(MAXMIND_LICENSE_KEY, 'maxm
             </table>
             </form>' . "\n";
 } else {
-    require_once __DIR__ . '/lib/request/Requests.php';
-    Requests::register_autoloader();
-
     ob_start();
     echo __('downfile15') . '<br>' . "\n";
 
-    $files_base_url = 'https://download.maxmind.com';
+    $urlSchema = 'https://';
+    $downloadServer = 'download.maxmind.com';
     $file['description'] = __('geoip15');
-    $file['path'] = '/app/geoip_download?edition_id=GeoLite2-Country&suffix=tar.gz&license_key=' . MAXMIND_LICENSE_KEY;
+    $file['path'] = '/geoip/databases/GeoLite2-Country/download?suffix=tar.gz';
     $file['destination'] = __DIR__ . '/temp/GeoLite2-Country.tar.gz';
     $file['destinationFileName'] = 'GeoLite2-Country.mmdb';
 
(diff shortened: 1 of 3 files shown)


mailscanner/geoip_update.php
```diff
@@ -80,44 +78,46 @@
...
-                if (USE_PROXY === true) {
+                $ch = curl_init($urlSchema . $downloadServer . $file['path']);
+                curl_setopt($ch, CURLOPT_HTTPAUTH, CURLAUTH_BASIC);
+                curl_setopt($ch, CURLOPT_FOLLOWLOCATION, true);
+                curl_setopt($ch, CURLOPT_USERNAME, MAXMIND_ACCOUNT_ID);
+                curl_setopt($ch, CURLOPT_PASSWORD, MAXMIND_LICENSE_KEY);
+                curl_setopt($ch, CURLOPT_USERAGENT, 'MailWatch/' . mailwatch_version());
+                if (defined('USE_PROXY') && USE_PROXY === true) {
+                    curl_setopt($ch, CURLOPT_PROXY, PROXY_SERVER);
+                    curl_setopt($ch, CURLOPT_PROXYPORT, PROXY_PORT);
```
endelwar, line 89: Did you try `Requests_Auth_Basic` in $requestSession->options['auth']? I don't like very much removing Requests integration

Skywalker-11: I

> Did you try `Requests_Auth_Basic` in $requestSession->options['auth']? I don't like very much removing Requests integration

To be honest I thought it might be good to remove the Request library as it is big piece of code for just downloading a file and it was orignally introduced for geoip download (if the commit history is correct) and isn't used anywhere else right now. But I'm fine if you want to keep it.

I did try the `auth` option of the requestsession but it didn't work that time. But I also didn't invest a lot of time trying. So it might just have been me 🤷.

endelwar: You are correct, `Requests` was introduced simplify the geoip file download in 2015, and it proved useful to manage automatically the redirection that MaxMind apply on the download link when they switched to CloudFlare R2 storage: with curl you have to check the first response for a 301/302 redirect location and then make a second request to download from the correct location. The actual library in `mailscanner/lib/request` has also been patched to support HTTP/2 calls.

Skywalker-11: The `CURLOPT_FOLLOWLOCATION` options tells curl to follow such redirects and I'm pretty sure that HTTP/2 is supported out of the box. Corresponding feature flags are supported since PHP 7.0.7 (release 05/2016) and matching curl versions also since 2016.

mailscanner/geoip_update.php
```diff
@@ -55,15 +55,13 @@
...
-    Requests::register_autoloader();
-
     ob_start();
     echo __('downfile15') . '<br>' . "\n";
 
-    $files_base_url = 'https://download.maxmind.com';
+    $urlSchema = 'https://';
+    $downloadServer = 'download.maxmind.com';
     $file['description'] = __('geoip15');
-    $file['path'] = '/app/geoip_download?edition_id=GeoLite2-Country&suffix=tar.gz&license_key=' . MAXMIND_LICENSE_KEY;
```
ghost, line 66: This is still working for my legacy Maxmind key, should we fall back and attempt this download when MAXMIND_ACCOUNT_ID is not populated?

mailscanner/geoip_update.php
endelwar: I know it seems anachronistic, but the 1.2.x branch needs to support PHP 5, so PHP 7+ solutions are a no go in 1.2 :(

We should start a 1.3 branch with php 7.4 support or even better align with officially supported only PHP version

PR #1316, opened 2025-09-18
Update SQLBlackWhiteList.pm to allow foo@*.domain.ext

riksoft:
Be able to blacklist like \*.domain.ext is great, but with this little change it also allows to set less broader rules like this one foo@*.spammer.com

Why it's useful? Example:
info@random1.domain.com  is OK
info@random2.domain.com  is OK
info@random3.domain.com  is OK
ads@random1.domain.com  is bad
ads@random2.domain.com  is bad

Can be solve blacklisting only ads@*.domain.com instead of *.domain.com that would be too much.


endelwar:
Thanks @riksoft !

riksoft:
Thanks to you guys!
(Salutami l'Italia).

PR #1321, opened 2026-03-17
Add Turkish Translation (tr.php)

alperperi:
Hello team, I have created the Turkish language file for MailWatch. Best regards.

*Added Turkish language support. Translated by Alper Peri.


endelwar:
Thanks for working on this translation.

I do **not** know Turkish myself, so this is only an **AI-assisted review** based on a comparison with `mailscanner/languages/en.php`, and the comments below should definitely be taken with some caution. That said, I did not notice syntax issues, missing keys, broken placeholders, or major mistranslations that would introduce clearly wrong meaning. 

A few items still look worth adjusting before merge:

- `messages05`: English is `Messages`, but Turkish is `Mesaj`. This should probably be `Mesajlar`.
- `spamwl04`, `spambl04`, `mcpwl04`, `mcpbl04`: current wording is understandable, but reads more like list names than status labels. It may be worth aligning them more closely with the “Allowlisted / Blocklisted” meaning of the English source.
- `title59`: `Karantina Serbest Bırakma` is understandable, but feels a bit literal. Something like `Karantinadan Serbest Bırakma` or `Karantinadan Çıkarma` may sound more natural.
- `unknownvirusscanner03`: the English source includes `(see FAQ)`, which seems to be omitted in Turkish.
- `message115`: the English source mentions that GeoIP files are updated on the first Tuesday of each month; that detail seems to be missing in Turkish. 

Also, beyond `tr.php`, it looks like Turkish should probably be added in the other translation files in the `// 99 - General` section near the end of those files, so the language list stays consistent. And `mailscanner/conf.php.example` should likely be updated as well, since it currently lists supported languages as `de, en, es-419, fr, it, ja, nl, pt_br` and the `USER_SELECTABLE_LANG` example does not include `tr`. 

So overall: this looks solid and useful, but I would suggest fixing `messages05`, considering the wording tweaks above, and updating the remaining language lists/config examples so Turkish is wired in consistently.

endelwar:
Thank you @alperperi !

alperperi:
Your welcome @endelwar, I hope it will be of some benefit to the community.

